Issue #47, opened 2017-10-15
show full transaction IDs on "Transactions" tab


djpnewton:
I am trying to check the status of my (on chain) transactions independently

bryanvu:
Fixed by PR #57.

Issue #22, opened 2017-10-12
Add Commmas to Large Numbers in Wallet


casesandberg:
https://twitter.com/lwsnbaker/status/918611275210297344

fresheneesz:
There are already commas in the bottom left, but yeah, would make sense to add commas to the numbers in "Your Wallet" in the Channels section.

bryanvu:
Fixed in PR #57

PR #57, opened 2017-10-18
Allowed displaying complete TxID and Pubkeys and Fixed commas and currency unit on wallet, channelist and transactionslist

DaReaper:
List of changes made:

1) Allowed displaying complete TxID on the transactionslist screen with word-wrapping for TxID and Notes when screen resizes. Added TxID and Notes text labels as well.

2) Allowed displaying complete Pubkey on the Wallets screen with word-wrapping when screen resizes. Added Pubkey text label as well.

3) Added commas to amounts in Wallet, Channel List and Transactions List with currency unit. This currency unit and amount display can probably be dynamically changed in the case where a future enhancement provides this facility from the 'settings' page

Fixes issues: https://github.com/lightninglabs/lightning-app/issues/47 and https://github.com/lightninglabs/lightning-app/issues/22


packages/lightning-core/transactions/TransactionsListItem.js
```diff
@@ -60,10 +57,10 @@ export const TransactionsListItem = ({ type, hash, memo, amount, status }) => {
       </div>
       <div style={{ ...styles.text, ...styles.column }}>
         <div style={{ ...styles.big, ...styles.overflow }}>
-          { hash }
+          <Text bold>TxID: </Text>{ hash }
         </div>
         { memo ? (
```
JimmyMow: This syntax below is probably cleaner if you don't have anything to return if the conditional is false :)

`{memo && <div style={{ ...styles.small, ...styles.overflow }}>Note: { memo }</div>}`

bryanvu:
Huge thanks for this @DaReaper! It's nice to not have to squint and count out zeros one by one. Before I merge this, was wondering if you could make two quick changes. First, on the channels screen, it looks like if I have 1 BTC or more, the SAT label will wrap down to the next line. See screenshot below. Is there a more graceful way to handle that?

screen shot 2017-10-22 at 9 52 14 pm

Also, could you use rebases instead of merge commits? Would ideally like to keep the history a little cleaner.

DaReaper:
@bryanvu I've fixed the currency unit from wrapping to next line on the Wallet and the Transactions screen and have rebased my branch like you've suggested. I hope it's good to merge now :)

bryanvu:
LGTM. Thanks @DaReaper and thanks @JimmyMow for the review!

Issue #21, opened 2017-10-12
Don't show QR code for receiving during initial sync


Roasbeef:
Deported by `djseeds` on slack:
```
I think the wallet QR code should probably be disabled during chain sync, because right now it generates a QR code of `bitcoin:Opening Wallet...`
```

Before the neutrino backend is fully synced, it won't be able to fetch an address from the RPC server as it doesn't start up until the chain backend has finished syncing. 

The send/receive tab itself should be shown as inactive, and should ignore user input until `neutrino` has fully synced.

Issue #24, opened 2017-10-13
Initializing Wallet / Opening Wallet...


fresheneesz:
I downloaded the windows version of this, ran the program, gave it network access when the popup came up. Next step is to send myself some sats from a faucet. I assume to do that I need to copy my address in the Receive section and send to it. Problem is that the address text line contains either "Opening Wallet..." or "Initializing Wallet" for like half an hour. Does it have to finish "Syncing to Chain" before I can use my wallet? I would have thought I could just have an address i could immediately use, even if the client couldn't see my transactions until its done syncing. 

Another problem, if I click the "Wallet Address" text, it copies the text in the address line (eg "Opening Wallet..."). Might want to disable copying the wallet address until there's an address to copy. 

The UI looks neat so far tho! Looking forward to testing it out once I can drink from a faucet!

DaReaper:
Prevent copying 'Opening Wallet', 'Initializing Wallet' values in Wallet Address field and disable QR code until address available.

![Screenshot of copying wallet address](https://user-images.githubusercontent.com/567661/31584418-d01edb9c-b1cb-11e7-9355-61d26814f6d2.png)
![Screenshot of scanned QR code](https://user-images.githubusercontent.com/567661/31584431-3aac758c-b1cc-11e7-9896-3e57cff9c8e4.png)

jcsmesquita:
Mine's also stuck on Initializing wallet in the Request tab. I was able to fund the wallet by grabbing the pub key in Settings though.

fresheneesz:
@tanx Could you please comment on why you closed this?

PR #58, opened 2017-10-18
Disable Payments and Payment Requests While Syncing

casesandberg:
Closes https://github.com/lightninglabs/lightning-app/issues/21 and https://github.com/lightninglabs/lightning-app/issues/24

* Inputs can be disabled
* QR Popup disabled while syncing
* Disable input copy to clipboard when disabled
* All GRPC calls guarded while syncing

### DIFF
--- a/packages/lightning-core/common/Input.js
+++ b/packages/lightning-core/common/Input.js
@@ -10,7 +10,7 @@ export class Input extends React.Component {
   handleBlur = () => this.setState({ focused: false })
   handleClick = () => {
     const { selectOnClick, copyOnClick, onCopy, onClick } = this.props
-    if (copyOnClick) {
+    if (copyOnClick && !this.props.disabled) {
       this.input.select()
 
       try {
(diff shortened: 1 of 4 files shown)


packages/lightning-core/common/Input.js
```diff
@@ -10,7 +10,7 @@ export class Input extends React.Component {
   handleBlur = () => this.setState({ focused: false })
   handleClick = () => {
     const { selectOnClick, copyOnClick, onCopy, onClick } = this.props
-    if (copyOnClick) {
+    if (copyOnClick && !this.props.disabled) {
```
JimmyMow: this is small but maybe wanna stay consistent and declare `disabled` on line 12 with the other props?

`const { selectOnClick, copyOnClick, onCopy, onClick, disabled } = this.props`

`if (copyOnClick && !disabled) {`

bryanvu:
@JimmyMow and @casesandberg are you guys seeing the QR code and Wallet Address copy to clipboard disabled when building with this PR? I don't see anything actually disabled during syncing when I build this.

JimmyMow:
hm @bryanvu I didn't actually build this PR, I just skimmed the changes. I can take a deeper look in a sec

Issue #18, opened 2017-10-11
Remove menu bar from Windows build


bryanvu:
In the windows build, the menu bar still appears with the "Lightning" and "Edit" drop-down menus. This is despite mainWindow.setMenu(null) in main.dev.js.

PR #59, opened 2017-10-18
Remove Menu Bar from Windows and Linux Build

casesandberg:
Closes https://github.com/lightninglabs/lightning-app/issues/18

### DIFF
--- a/apps/desktop/main.dev.js
+++ b/apps/desktop/main.dev.js
@@ -1,4 +1,4 @@
-/* eslint-disable global-require, no-console */
+/* eslint-disable */
 
 import { app, BrowserWindow, Menu } from 'electron'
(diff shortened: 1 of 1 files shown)


bryanvu:
Thanks @casesandberg! 

How is this change different than the commented out lines here: https://github.com/casesandberg/lightning-app/blob/4840376b8e2077d343d2a44601a6e7dd4751053e/apps/desktop/main.dev.js#L216. Also, previously, I added this line: https://github.com/casesandberg/lightning-app/blob/4840376b8e2077d343d2a44601a6e7dd4751053e/apps/desktop/main.dev.js#L139, which removed the menu on the Mac version. Is this still necessary with this latest change?

Issue #38, opened 2017-10-15
Wallet balance doesn't auto-update dynamically after completion of chain sync


DaReaper:
The Lightning app's wallet balance displays as '0' after every sync completes. The ideal result would be to auto-update and display the correct balance after the sync completes instead of display it as '0'. The user currently has to click on the balance or 'Channel' screen (for example pay, request etc) in order to view the updated balance.

0 balance after sync:
![lightning0bal](https://user-images.githubusercontent.com/567661/31584610-2dadf9e8-b1cf-11e7-96ec-54bbee015d33.png)

Balance updated on clicking 'Channel':
![balanceupdatedchannel](https://user-images.githubusercontent.com/567661/31584614-3acad4ac-b1cf-11e7-9399-671c7059961a.png)

PR #60, opened 2017-10-18
Fetch Balance when Fetching Accounts

casesandberg:
Fixes https://github.com/lightninglabs/lightning-app/issues/38


bryanvu:
Thanks @casesandberg! Verified that this behavior was happening and that the balance was only updated when you clicked on the Channels page after the sync completed, rather than automatically when the sync completed. Also verified the fix.

PR #74, opened 2017-10-22
Add the key 'command' to ps.lookup

### LINKED ISSUE #31: lightning-app dislikes certain usernames
# Problem

v0.1-alpha will fail to run for a user that contains 'lnd' within the username. Possibly also 'lnd' within the hostname (originally I tested lnd@lnd, lnd2@lnd2, lnd3@lnd3...).
Tested on Ubuntu 16.04, 17.04, 17.10.
The error message `lnd Already Running` prevents the interface from loading and `lnd` never appears in `ps`.

# Solution
The immediate workaround for me was to create a new VM with a different username/hostname not containing `lnd`.
I believe the culprit to be the `ps.lookup` function in `apps/desktop/main.dev.js`. I suspect that it's doing somethine like grepping the `ps` output, seeing `lnd` in a directory name or user name and assuming it's running. Not quite sure.

This was a bit of a wild goose chase :)

PowerInside:
This should fix #31

### DIFF
--- a/apps/desktop/main.dev.js
+++ b/apps/desktop/main.dev.js
@@ -20,7 +20,7 @@ const isDev = process.env.NODE_ENV === 'development'
 const runningProcesses = []
 
 const isProcessRunning = command => new Promise((resolve, reject) => {
-  ps.lookup({ command },
+  ps.lookup({ command: command },
     (err, resultList) => {
       if (err) { throw new Error(err) }
       resultList[0] ? resolve(resultList[0]) : reject()


PowerInside:
I've tried running by changing to the lnd user from my machine itself (also tried setting my hostname to lnd) and it works there. I'll try this under a docker container.

JimmyMow:
`ps.lookup({ command }`

is actually leveraging `es6 Enhanced Object Literals`. the result of the above code is:

`ps.lookup({ command: command }` 

so this PR actually isn't adding any functionality :)

as far as #31, that is an interesting and tricky bug, and I don't have any ideas off the top of my head to potentially solve unfortunately ☹️

esotericnonsense:
Hello, I authored the original issue.
Yes, we're using property shorthand there.
I don't have much knowledge of the nodejs ecosystem, I'm vaguely wondering whether ps.lookup can handle this at all.

JimmyMow:
Well the results given in the callback have some additional info such as `command`, `arguments`, `pid` and `ppid`. Could easily implement some logic that checks for `LND` arguments to the process that was found. That feels a bit hacky imo and there should be an easier way

PowerInside:
Oops. Wasn't aware of that es6 feature. Closing. Sorry about that!

PR #80, opened 2017-10-25
Revert "Fix "blank app on start" bug"

bryanvu:
Reverts lightninglabs/lightning-app#77. As noted by @DaReaper, this change breaks startup in dev mode on Windows.


thejourneydude:
How do we effectively test all three platforms so my mistake doesn't happen again? Many developers don't have all three OS platforms to test on and this can be time intensive. I'll do some research some more and think about this.

Issue #96, opened 2017-11-22
In dev mode, the App should first look for lnd in the PATH


bryanvu:
### Description

Currently, the App is hard-coded to start using the lnd that's bundled in the apps/desktop/bin directories. If there is a version of lnd already existing in the PATH, the App should use that lnd rather than the bundled lnd.

### Steps to reproduce the behavior

Start the App with lnd in the PATH.

### Expected behavior

The App will start running with the lnd in the PATH.

### Actual behavior

The App will start running with the bundled lnd.

PR #101, opened 2017-12-06
Adding the possibility to run an existing lnd if it's on PATH.

diofeher:
Related to #96


bryanvu:
Thanks for this @diofeher! I was attempting to test this, and I added a "dummy" lnd in my path, which was just a blank file. I was expecting that when I ran this build since the dummy lnd would precede the bundled lnd, and I would get some kind of error, but it seemed to just use the lnd that was packaged with the app. Any thoughts on why that would be happening?

yate:
It's because `type` isn't an executable that is in PATH, it's a built in shell command, so spawnSync won't be able to execute it.  My suggestion would be to try to spawn lnd and if there are errors, spawn the packaged version.

diofeher:
So what I've tried to do is to check if lnd was in the PATH, and if yes, spawn the command.

I was playing around on how you tested @bryanvu . If I add an empty executable, no errors are raised. What we could do is to check the output of lnd --version and if returns something valid, we can continue to spawn the process. I'm doing that way now, will send a PR later.

bryanvu:
Thanks @diofeher. I think I understand what the new version is doing, but it seems to me that it won't work on Windows. On Windows, I believe the executable name won't match with the beginning of the version string. Also, it seems like what's going on is that if an invalid lnd is in the path, then it will automatically skip and use the packaged lnd.

IMO, if there's an invalid lnd in the path, I think it would make more sense to just throw an error and quit startup. If we just skip to to the packaged lnd, if the user has an invalid lnd for some reason, that could cause confusion since it won't be obvious that they're running the packaged lnd.

diofeher:
My only question is on how to detect if the binary execution is invalid. For example:

```
touch lnd
chmod +x lnd
```
then on node.js:
```
> cp = require('child_process')
> ret = cp.spawn('/Users/diofeher/workspace/go_projects/bin/lnd2')
> ret.pid
87595
```

The executable was spawned as valid, even without containing any code. What do you guys suggest on that?

diofeher:
@tanx any reason to close it?

tanx:
@diofeher this one is already working on the current master

diofeher:
Ok, awesome!

Issue #54, opened 2017-10-17
Add a console section to the settings page


Roasbeef:
A "console" page should be added to the settings page. This would let users manually enter `lncli` commands like adding peers, listing their channels, changing the debuglevel, etc. 

Such a page would also aide in debugging as users will be able to input directed commands in order to get a better feel for what the issue at hand is.

PR #104, opened 2017-12-08
Adding lncli access to Settings tab

diofeher:
Related to #54 

This pull request is to see if I'm heading to the right direction. The interface should be used to interact **only** with lncli?

What about creating a separate tab on the left menu to interact with the CLI (separating from settings, doesn't make sense to put it there, or at least rename Settings to "Advanced Use").


diofeher:
screen shot 2017-12-08 at 10 16 57

diofeher:
If **lncli** is the only command necessary to be run here, we can remove the dependency of react-bash.

bryanvu:
Thanks @diofeher! I'm going to do a bit more testing but we only need lncli, so I would recommend removing the option for any other commands.

bryanvu:
Oh, and about your other suggestion, I think that renaming settings to "Advanced" makes sense. I think that tab could contain the Logs and the Console. We can take out the "Settings" for now and add it back in later.

diofeher:
Thanks for the comments @bryanvu !

Here's the updated UI:

screen shot 2017-12-11 at 16 52 38

diofeher:
I've added the ability to see the command you sent on the logs and I've added a timestamp.

screen shot 2017-12-12 at 11 37 36

diofeher:
@bryanvu anything missing to merge this?

bryanvu:
@diofeher, huge apologies for the slow response on this. Am just recovering from the holidays and work on a couple other projects. I just pulled this and built it (npm run package-electron), and when I enter anything in the "lncli" text box, I'm not getting any additional output in the console. I tried "getinfo" and "help". However, when I run in dev mode (npm start), it does work. Any thoughts on why it doesn't work with the packaged version?

diofeher:
@bryanvu I need one favor, to add lncli binaries to windows and linux :)

diofeher:
@tanx any reason to close it?

tanx:
@diofeher sorry I don’t even remember closing this.

Issue #42, opened 2017-10-15
Generate / Display QR code when generating a Request for lightning payment and ability to scan QR code when making a Payment request


DaReaper:
It would be useful for a POS operator to be able to display a Request for payment QR code to a customer when using the application. Hence, if a QR code is generated for a lightning payment request just like the Bitcoin receiving address QR code it would be very useful.

Also, at the same time like on the https://starblocks.acinq.co webpage as soon as you make a purchase it displays the payment request code on the checkout page with a QR code to scan. It would be useful to provide the ability to provide a scan button to invoke the webcam / scanner to scan a QR code when making payments.

Note: This isn't an issue, but a feature request.

PR #107, opened 2017-12-10
Display a QR code on payment request modal

yate:
part of #42


yate:
Replacing lightning:// with lightning: in the payment request allowed me to scan it with eclair mobile wallet. Should the real fix be in the payment request encoding? https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#encoding-overview

bryanvu:
Thanks for this @yate! FYI I'm going to add a quick followup commit to this to replace "lightning://" with "lightning:" in a simpler way as proposed by @Roasbeef for PR #113.

Issue #95, opened 2017-11-16
When App starts in dev mode, pass command-line parameters to lnd


bryanvu:
### Description

Currently, when the App starts up in dev mode, the configuration is read from lnd.conf. This makes it difficult to run two instances of the app on the same machine, which would be useful for testing and such. the `npm run start` command should take command-line arguments for port numbers (and other configuration options) and use those when starting lnd.

### Expected behavior

`npm run setup --peerport=10019 --rpcport=10017 --datadir=test_lnd --logdir=test_lnd` should start up the App with a second instance of lnd using a different set of ports and file locations.

t4sk:
@bryanvu 
"`npm run setup ...` should start up the App"

Reading the description, I think you meant to say
"`npm start --peerport=10019 --rpcport=10017 --datadir=test_lnd --logdir=test_lnd` should start up the App".

diofeher:
Started to work on that.

PR #108, opened 2017-12-13
Adding the ability to pass arguments in npm start to lnd command

diofeher:
The only workaround that I had to do is that you have to write extra -- to pass arguments, like:

npm run start -- --peerport=10019 --rpcport=10017 --datadir=test_lnd --logdir=test_lnd

Because that's the way npm can pass custom arguments in its scripts.

screen shot 2017-12-12 at 21 03 52

Related to #95

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,11 @@ You can run the app in development mode in lieu of using the packaged app:
 npm start
 ```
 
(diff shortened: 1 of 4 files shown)


bryanvu:
Thanks @diofeher. Nice work!

Issue #109, opened 2017-12-13
Varying amount in lightning payment has no affect


anaoum:
### Description

When pasting a payment request, the amount field is pre-populated with the amount from the payment request. However, this amount can then be modified in the user interface. Once submitting this payment, checking it in the transactions tab shows the original amount from the payment request.

### Steps to reproduce the behavior

1. Enter payment request on the payment tab.
2. Change the amount that was pre-populated after the above step.
3. Send payment.
4. Check the transactions page tab for the previous transaction.

### Expected behavior

Either the amount should not be editable for lightning payments (not sure what the protocol dictates here), or the altered amount should be the amount sent in the transaction.

### Actual behavior

The transaction sends the amount as specified by the encoded payment request, regardless of the amount entered by the user.

PR #112, opened 2017-12-18
Disable amount field for invoice payment requests

bulenttastan:
As described in [this issue](https://github.com/lightninglabs/lightning-app/issues/109) if you're using an invoice address the user shouldn't not be able to change the amount field. Changing the amount does not change the actual amount sent, so it's confusing to leave that field enabled.

This PR disables the field if an invoice is used, otherwise leaves it enabled.


bryanvu:
Thanks @bulenttastan! LGTM.

yate:
Maybe this should only be disabled if there is an amount included on the payment request, otherwise we wouldn't be able to send payments to an invoice that doesn't specify the amount.

bulenttastan:
@yate I see your point but sometimes you fill out the form bottom up, like the amount first then the address. I thought about doing it the way you described, but figured it would be better to disable only if needed.

PR #113, opened 2017-12-20
Better sanitizePaymentRequest

BeeCodiing:
Really simple but functional. I noticed that starblocks's payment invoice uses `lightning:` and not `lightning://` (the only one specified by the function) so I added support for that prefix too by using regex.

### DIFF
--- a/packages/lightning-core/helpers/paymentRequest.js
+++ b/packages/lightning-core/helpers/paymentRequest.js
@@ -5,5 +5,5 @@ export const decoratePaymentRequest = (pr) => {
 }
 
 export const sanitizePaymentRequest = (pr) => {
-  return pr.replace(prefix, '')
+  return pr.replace(/lightning:(\/\/)?/g, '')
 }
--- a/packages/lightning-core/helpers/paymentRequest.js
+++ b/packages/lightning-core/helpers/paymentRequest.js
@@ -4,6 +4,7 @@ export const decoratePaymentRequest = (pr) => {
   return prefix + pr
 }
 
+// Payment requests should start with either lightning: or lightning://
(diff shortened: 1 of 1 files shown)


packages/lightning-core/helpers/paymentRequest.js
```diff
@@ -5,5 +5,5 @@ export const decoratePaymentRequest = (pr) => {
 }
 
 export const sanitizePaymentRequest = (pr) => {
-  return pr.replace(prefix, '')
+  return pr.replace(/lightning:(\/\/)?/g, '')
```
bulenttastan: Can you add comment explaining the possible prefix matches?

BeeCodiing: I just found the two above (`lightning:` and `lightning://`) so they are the only possible matches

bulenttastan: I meant to add it as a comment in the code above that line, something like:
 `// Payment requests should start with either lightning: or lightning://`

BeeCodiing: oh sorry for my misunderstanding. I'll do it rn

yate: Is there any documentation for the `lightning://` prefix? I could only find https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md which says not to use `lightning://`

Roasbeef: Our usage of the `lightning://` prefix actually predates BOLT 11 itself.

bryanvu:
Thanks @BeeCodiing and @bulenttastan! This was completed as a followup to PR #107.

PR #148, opened 2018-01-18
adding storybook to lightning-components

kaplanmaxe:
It seems as if someone had tried adding storybook to `lightning-components`, however did not finish the configuration. There were multiple `story.js` files in the repo, however no configuration was set. This PR adds storybook to `lightning-components` which will greatly ease development.

Note: Some stories are throwing react errors. Maybe should be fixed in a separate PR?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ main.js.map
 stats.json
 
 test_lnd
+.vscode/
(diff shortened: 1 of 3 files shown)


bryanvu:
Thanks @kaplanmaxe. I was wondering what this storybook stuff was all about. It seems like it kind of works for lightning-components (with a bunch of errors as you mention), but this is an improvement over where we were before.

Issue #138, opened 2018-01-14
Node public key plus IP is not shown in application


MidnightLightning:
### Description
When creating a channel to another Lightning Network node, the syntax needed is "Pubkey@HostIP". For telling friends whom I want to transact with what my node is, the application shows my pubkey on the "Channels" page, but doesn't show my IP anywhere. So, if I want to share my connection address with someone, I need to figure out my own IP address through some other means.

### Steps to reproduce the behavior
Go to the "Channels" page and note the pubkey

### Expected behavior
Have somewhere in the application a copy/paste-able string to send to someone else who wishes to open a channel with me.

### Actual behavior
Only half the information needed is shown in-app and the user is responsible for figuring out their external IP address.

PR #154, opened 2018-01-19
adding IP as part of public key

kaplanmaxe:
Implements #138 

Using ipify.org which is a popular public API without authentication / limits.

### DIFF
--- a/apps/desktop/main.dev.js
+++ b/apps/desktop/main.dev.js
@@ -13,7 +13,6 @@ import fs from 'fs'
 import grpc from 'grpc'
 
 app.commandLine.appendSwitch('remote-debugging-port', '9997')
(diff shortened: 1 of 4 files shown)


kaplanmaxe:
@tanx just an FYI this shouldn't be labeled v2 as it is against the master branch

tanx:
@kaplanmaxe ooops. Thanks for the heads up.

bryanvu:
@kaplanmaxe, I discussed this with @Roasbeef, and we decided that since we don't have NAT traversal, this could be confusing to folks who may try to publish non-working IP addresses for others to connect to. There's also potentially a potential privacy leak to the third-party IP detection service, which may be able to identify possible lightning nodes that were intended to be private. I'm going to close for now, as I think this requires a bit more discussion.

Issue #152, opened 2018-01-19
MacOS - "Hide" menubar commands


photonclock:
### Description
GUI feature for MacOS:
Add "Hide", "Hide Others", "Show All" commands to "Lightning" menubar of MacOS version of Lightning App

### Steps to reproduce the behavior
Press command-H or command-option-H

### Expected behavior
command-H or select "Hide" from "Lightning" menu - hide Lightning app
command-option-H or select "Hide Others" from "Lightning" menu - hide other applications leaving Lightning app in foreground

### Actual behavior
Hide and Hide Others commands are not implemented in Lightning menu.
"Show All" command is a related command that is considered universal for a MacOS application's menubar.

bryanvu:
Addressed by PR #157.

PR #157, opened 2018-01-20
Add hide, hide others, and show all menu options

kaplanmaxe:
Implements #152


bryanvu:
Thanks @kaplanmaxe! Super nitpicky request, but I see in my other apps, that there's a separator line between these three options and the "Quit" option. Would it be possible to add that? Also, it looks like in order to be consistent, that "Quit" should be changed to "Quit Lightning".

kaplanmaxe:
@bryanvu good call! I have updated per your comment.

bryanvu:
LGTM. Thanks @kaplanmaxe!

Issue #139, opened 2018-01-14
Lock down the "Create Channel" form after submit button is pushed


MidnightLightning:
### Description
When a user clicks on the "Create Channel" link in the "Channels" tab, they're taken to a form to create a new channel. Initially, the submit button is grayed out and disabled. If a user puts in a pubkey and IP address and an amount, the button gets enabled.

When you press the submit button, there's a delay, and then the user is taken to the "Channels" tab again, where the newly-created channel is now listed, as "Pending-open". The issue is that the delay between pressing the submit button and being taken to the Channels page is noticeable (a second or two), but in that time there's no indication the app is doing anything (no "loading..." indicator), and the form itself isn't locked down (the submit button doesn't get disabled again).

### Steps to reproduce the behavior
Go to Create Channel form and submit the form.

### Expected behavior
Form locks until the page changes, and gives some indication the application is doing something.

### Actual behavior
No visual change from pressing the submit button until after a few seconds pass.

r0zar:
I've tested and can confirm this issue on v0.1.4-alpha. It's very odd and concerning as if you navigate away from the page before the redirection, the transaction will go through but the channel will not show as pending-open.

r0zar:
The 'opening channel' alert should show immediately after pressing the button, opposed to after being routed to the channels page.

machine-uprising:
Agree this is needed. A notification "opening channel" that shows immediately after pressing the button to know that execution is occurring will help.

PR #158, opened 2018-01-21
Show 'opening channel' alert directly after user submits form

kaplanmaxe:
Related to #139 

Show 'opening channel' alert directly after user submits form to indicate something is happening.


bryanvu:
Hi @kaplanmaxe, not sure what's going on with this, but when I attempt to open a channel, it does show the "Opening Channel" toast, but then it flashes up some kind of very long toast (with an error of some kind I assume), and then the "Opening Channel" toast goes away just as quickly as it does normally. This is when testing on Simnet. Any idea what's going on?

Issue #125, opened 2018-01-01
Show version somewhere in the app


copumpkin:
I wanted to report a bug and couldn't find the version of the app I'm running anywhere, even in the log panel.

bryanvu:
@copumpkin, this is a good feature request that we should get in to an "About" dialog or something at some point. In the meantime, if you're using one of the binaries from the releases page, the version number is in the filename. If you've built the app, you can use the commit hash.

machine-uprising:
Could this be added to the settings page? Might reduce the need for an extra "About" dialog

robertDurst:
@copumpkin @machine-uprising found this to be a pretty easy addition. Here are a couple options I came up with for placement on the "Settings" page. Let me know your thoughts!

## Option 1
<img width="1440" alt="screen shot 2018-01-20 at 10 15 29 pm" src="https://user-images.githubusercontent.com/16689974/35191584-f7ac5ff4-fe33-11e7-9a05-f74c123bc34f.png">

## Option 2
<img width="1440" alt="screen shot 2018-01-20 at 10 46 21 pm" src="https://user-images.githubusercontent.com/16689974/35191585-fd44f502-fe33-11e7-875c-1822c143a4c0.png">

robertDurst:
## Option 3 (Bottom Left -- Sidebar)
<img width="1440" alt="screen shot 2018-01-21 at 12 22 48 pm" src="https://user-images.githubusercontent.com/16689974/35198591-e10a7188-fea5-11e7-9666-3948cdcc4c53.png">

machine-uprising:
@robertDurst - I think I prefer the App Version being on the settings page instead of in the side bar. 
A question I also have about it being in the side bar is what happens when the "Syncing to Chain" message is shown in that location (see attached image showing this message). 

Having it on the Settings tab makes more sense to me. It's not something I would need to see all the time but is good to have easily available when you need it. So to take up real estate on the side menu bar might not be neccessary. 

I like it next to the Settings heading as per your Option 2. 

![issue 125 - showing app version - syncing to chain](https://user-images.githubusercontent.com/32746160/35457142-ae0d9b2e-02b6-11e8-8291-4bf448a28970.png)

robertDurst:
@machine-uprising thanks for the feedback! When it says "Syncing to chain" the version will just appear below this message, "rooted" to the bottom of the sidebar.

[...]

PR #161, opened 2018-01-22
Add app version to the bottom of the sidebar

robertDurst:
Added app version to the bottom of the sidebar for easy reference. Reference #125

### Complete Page View
screen shot 2018-01-21 at 7 00 10 pm

### Close Up
screen shot 2018-01-21 at 7 01 20 pm


bryanvu:
Hey @robertDurst, thanks for this! I tested and it works, but it looks a bit cramped while the "Syncing to Chain" notifier is up, and I think the left bar looks a bit cluttered anyway. I think the version number should be in an "About" menu or something similar. I suggest you, me and @tanx figure out where to put it for v2.

robertDurst:
Sounds good. Very simple addition, more of a design decision than any real coding.

PR #175, opened 2018-01-27
adding payment request descriptions if present and tests for payments

kaplanmaxe:
Add description for payment requests if present when paying and started test coverage for payments


.gitignore
```diff
@@ -24,4 +24,5 @@ package-lock.json
 /dist
 
 /lnd_data
+.vscode/
```
tanx: Please use a global gitignore for your machine, as everyone uses a different editor.

src/actions/payments.js
```diff
@@ -75,8 +75,7 @@ class ActionsPayments {
           pay_req: paymentRequest,
         })
         .then(response => {
-          resolve(response.num_satoshis);
-          // resolve(response);
+          resolve(response);
```
tanx: What's the reasoning behind changing the return value?

src/views/pay.js
```diff
@@ -34,19 +37,29 @@ class Pay extends Component {
...
             actionsPayments
               .decodePaymentRequest(payment)
-              .then(num_satoshis => this.setState({ amount: num_satoshis }))
-              .catch(() => {});
+              .then(res => {
+                this.setState({
+                  paymentRequest: true,
+                  amount: res.num_satoshis,
+                  description: res.description,
+                });
```
tanx: In general we should be using the mobx store instead of component specific state (see channels view). I haven't had a chance to go though all components yet. @kevinejohn is there a reason state is being used in some places and the mobx store in others?

tanx:
@kaplanmaxe thanks for the PR :) I made some comments. So glad we have a unit test infrastructure in place now for tests now. Great to see you've added some!

test/unit/actions/payments.spec.js
```diff
@@ -0,0 +1,48 @@
...
+        fallback_addr: '',
+        cltv_expiry: '9',
+      });
+      let body;
+      try {
+        body = await actionsPayments.decodePaymentRequest('goodPaymentRequest');
+      } catch (e) {
+        // Fail the test
+        expect(1 + 1, 'to be', 3);
+      }
```
tanx: You don't need this try/catch. If the `actionsPayments.decodePaymentRequest` throws an exception, the test will fail anyway.

src/actions/payments.js
kaplanmaxe: To be able to access both `num_satoshis` and `description` from the view

src/views/pay.js
kaplanmaxe: I will leave it as is for now. If you want me to change it to use mobx, lmk or it could probably be a separate PR. Have never used it before, but could play around with it.

kevinejohn: In general I only use mobx for shared global state and setState for anything local to a specific component. It keeps things more modular.

src/actions/payments.js
tanx: Ok. We should probably stay consistent with the return values in other actions and map to camel case like in `action/channels.js` In that case.

src/views/pay.js
tanx: The current tests strategy is to unit test the mobx actions, computed logic which let‘s us test the store as well. I‘d prefer to keep the react components purely functional and stateless if possible to improve testability and consistency. So I‘d prefer to move state changes to actions/computed code.

kevinejohn: That sounds good to me.

tanx: 👍🏼

kaplanmaxe:
I've refactored `decodePaymentRequest` to make use of mobx and updated the views accordingly. Test coverage added.

src/actions/payments.js
```diff
@@ -66,22 +66,18 @@ class ActionsPayments {
...
-        .catch(err => {
-          reject(err);
-        });
-    });
+    paymentRequest = paymentRequest.replace(PREFIX_URI, ''); // Remove URI prefix if it exists
+    try {
+      const request = await this._actionsGrpc.sendCommand('decodePayReq', {
+        pay_req: paymentRequest,
+      });
+      this._store.paymentsResponse = request;
```
tanx: Love the use of async/await 👍For consistency's sake we should map the comand's [response attributes](http://api.lightning.community/#decodepayreq) to camelCase attributes like in the [channel actions](https://github.com/lightninglabs/lightning-app/blob/abeb8839b2dbf382fb17b459609c232c3f206447/src/actions/channels.js#L36-L45). It's a bit more boilerplate code but readers of the code won't have to look what the return values are in the lnd api docs.

src/views/pay.js
```diff
@@ -19,6 +21,8 @@ class Pay extends Component {
 
   render() {
     const { payment, amount } = this.state;
```
tanx: Out of scope for this PR ... but we should remember to refactor the rest of the view to use the mobx store. I created a ticket: https://github.com/lightninglabs/lightning-app/issues/192

test/unit/actions/payments.spec.js
```diff
@@ -0,0 +1,42 @@
+import { observable, useStrict } from 'mobx';
+import ActionsGrpc from '../../../src/actions/grpc';
+import ActionsPayments from '../../../src/actions/payments';
+
+describe('Payments Unit Tests', () => {
```
tanx: `Actions Payments Unit Tests` here.

src/computed/payments.js
```diff
@@ -0,0 +1,18 @@
...
+  extendObservable(store, {
+    computedPaymentRequest: computed(() => {
+      const { paymentsResponse } = store;
+      if (!paymentsResponse) {
+        return {};
+      }
+      return {
+        numSatoshis: paymentsResponse.num_satoshis,
+        description: paymentsResponse.description,
+      };
```
tanx: As mentioned above, this mapping is done in the action for the other channels view. We should stay consistent. So I'm not sure we even need a computed function, since we aren't deriving new values in this case.

test/unit/computed/payments.spec.js
```diff
@@ -0,0 +1,37 @@
...
+          cltv_expiry: '9',
+        },
+      });
+      ComputedPayments(store);
+      expect(
+        store.computedPaymentRequest.description,
+        'to be',
+        '1 Espresso Coin Panna'
+      );
+      expect(store.computedPaymentRequest.numSatoshis, 'to be', '1700');
```
tanx: Love the test to map the lnd api return values to the expected local camel case syntax. As mentioned above, this should be mapped/tested in the action.

src/computed/payments.js
kaplanmaxe: I was wondering whether I needed a computed function or not. It seemed like that was the pattern you were following, so I added it, however I am new to mobx. Would I be better off just pulling from `paymentsResponse` in the view? Or should I add a new property to the store called something like `paymentRequest`?

tanx: We used the `*Response` naming in the other actions, but `paymentRequestReponse` just reminds me of Java's `FooFactoryImplAdapter` :D But `paymentRequestReponse` is probably correct in terms of consistency's sake.

kaplanmaxe:
Refactored based on last review. Removed computed function, mapping properties in action.

src/views/pay.js
kaplanmaxe: Agreed, I can handle this ticket.

src/store.js
```diff
@@ -24,7 +24,7 @@ class Store {
       paymentsResponse: null,
       peersResponse: null,
       channelsResponse: null,
-
+      paymentRequestResponse: observable({}),
```
tanx: Are you sure we need to declare a separate observable here? The store instance `this` is already declared as an observable [here](https://github.com/lightninglabs/lightning-app/pull/175/files#diff-3bf35cd41da51153dcca3003656f7c67R10).

kaplanmaxe: Nope! Good catch, I removed it. I'm kind of wrapping my head around observables / mobx as we go here, so appreciate you pointing out things like this! I wrapped it in `observable` as the property below it was, however I tested it out and still worked as expected after I removed it.

tanx:
Looks good! Thanks @kaplanmaxe

PR #179, opened 2018-01-28
adding ip, updating layout of channel header, adding test coverage

kaplanmaxe:
Implements #138 & v2 refactor of #154 

Adding IP to the UI, layout changes in `ChannelsHeader`, starting test coverage for wallet.

Using ipify.org which is a popular, public API without authentication / limits.


src/actions/wallet.js
```diff
@@ -97,6 +97,16 @@ class ActionsWallet {
...
   }
+
+  async getIP() {
+    try {
+      const request = await fetch('https://api.ipify.org?format=json');
+      const response = await request.json();
+      this._store.IP = response.ip;
+    } catch (e) {
+      throw new Error('Error fetching IP');
+    }
```
tanx: Seems like the app would crash if IP detection via the api fails. Either we log `e` to the console or handle it some other way.

src/actions/wallet.js
```diff
@@ -97,6 +97,16 @@ class ActionsWallet {
         this.t2342 = setTimeout(() => this.getNewAddress(), RETRY_DELAY);
       });
   }
+
+  async getIP() {
+    try {
+      const request = await fetch('https://api.ipify.org?format=json');
+      const response = await request.json();
+      this._store.IP = response.ip;
```
tanx: Please rename to `store.ipAddress`.

src/actions/wallet.js
```diff
@@ -97,6 +97,16 @@ class ActionsWallet {
         this.t2342 = setTimeout(() => this.getNewAddress(), RETRY_DELAY);
       });
   }
+
+  async getIP() {
```
tanx: Please rename to `getIPAddress`.

kaplanmaxe:
@tanx updated per review

tanx:
Thanks @kaplanmaxe. Right now it's fine to detect the IP. We should deactivate this feature by default thought to protect the user's privacy. Created a ticket here: https://github.com/lightninglabs/lightning-app/issues/197

Issue #193, opened 2018-01-30
Package crossplatform lnd build in Github release using travis


tanx:
We want to be able to release `lnd` and `btcd` binaries together with the lightning-app build for windows, mac os, and linux via Travis CI and publish the bundles to the [github repos releases](https://github.com/lightninglabs/lightning-app/releases).

tanx:
@valentinewallace expressed interest to tackle this issue. I can’t assign the github issue to her, since ahe isn’t in the lightninglabs github org. @roasbeef perhaps we can create a `Contributors` team that has write access to issues?

tanx:
Thank you @valentinewallace

PR #232, opened 2018-03-08
Automate multi-platform builds for releases via travis

valentinewallace:
Solves #193, resulting in a release that looks like [this](https://github.com/valentinewallace/lightning-app/releases/tag/v2_automate_releases). Manual parts are still signing the hashes file + adding release notes. 

For this to work, an api_key for travis needs to be generated from the travis CLI and put into this file, plus the branch name if necessary (defaults to master). :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,45 +1,38 @@
 language: node_js
 node_js:
-- '9'
+  - '9'
 
 services:
-- docker
+  - docker
 
 before_install:
-- ./assets/script/install_lnd.sh
+  - ./assets/script/install_lnd.sh
 
 before_cache:
-- rm -rf $HOME/.cache/electron-builder/wine
+  - rm -rf $HOME/.cache/electron-builder/wine
 
 cache:
   directories:
-  - node_modules
-  - $HOME/.cache/electron
-  - $HOME/.cache/electron-builder
+    - node_modules
+    - $HOME/.cache/electron
+    - $HOME/.cache/electron-builder
 
 env:
-- NAP_TIME=20000
+  - NAP_TIME=20000
 
 before_deploy:
-- env | grep -iE 'DEBUG|NODE_|ELECTRON_|YARN_|NPM_|CI|CIRCLE|TRAVIS_TAG|TRAVIS|TRAVIS_REPO_|TRAVIS_BUILD_|TRAVIS_BRANCH|TRAVIS_PULL_REQUEST_|APPVEYOR_|CSC_|GH_|GITHUB_|BT_|AWS_|STRIP|BUILD_' > env.txt
-- docker run --rm --env-file env.txt --env ELECTRON_CACHE="/root/.cache/electron" --env ELECTRON_BUILDER_CACHE="/root/.cache/electron-builder" -v ${PWD}:/project -v ${PWD##*/}-node-modules:/project/node_modules -v ~/.cache/electron:/root/.cache/electron -v ~/.cache/electron-builder:/root/.cache/electron-builder electronuserland/builder:wine /bin/bash -c "npm i; npm run build; react-scripts build; ./node_modules/.bin/electron-builder --em.main=build/electron.js --linux tar.gz --win zip --mac zip"
-- rm env.txt
-- PACKAGE_VERSION=$(node -pe "require('./package.json').version")
-- cd dist
-- shasum -a 256 Lightning* | sudo tee manifest-v${PACKAGE_VERSION}.txt
-- sudo chown -R travis:travis ./
-- cd ..
+  - ./assets/script/build_lnd.sh
 
 deploy:
   provider: releases
   api_key:
-    secure: <travis-cli generated api key>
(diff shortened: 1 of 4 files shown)


tanx:
@valentinewallace awesome! Thanks for the PR. I'll take a look asap.

.travis.yml
```diff
@@ -1,15 +1,45 @@
...
+- cd ..
+
+deploy:
+  provider: releases
+  api_key:
+    secure: <travis-cli generated api key>
+  file_glob: true
+  file:
+    - dist/Lightning*
+    - dist/manifest*
```
tanx: I don't have strong feelings regarding indentation in the .yml file. We should try to keep things consistent though. Preferably there'd be 2 spaces in front of every `-` like there is here.

tanx:
> Solves #193,

Just FYI. Not sure this keyword works: https://help.github.com/articles/closing-issues-using-keywords/

I just use:
Closes #193

valentinewallace:
Thank you, will integrate changes soon!! Also TODO: add the lnd+btcd binaries to the builds (presuming this is still desired). Forgot xD

tanx:
> Also TODO: add the lnd+btcd binaries to the builds (presuming this is still desired). Forgot xD

Good catch. I missed that. lnd+btcd are already being built using a shell script (at least for linux): https://github.com/lightninglabs/lightning-app/blob/28f0cb1b9b1226c11ef73a61d51f235dcb5663f0/.travis.yml#L12

Ideally we'd detect the OS in the shell script. See: https://stackoverflow.com/questions/394230/how-to-detect-the-os-from-a-bash-script

Then and bundle the binary from `/assets/bin/<os>` respectively.

Roasbeef:
Nice work! This is so cool 😎

Does it always publish as a final release, or is it possible to publish as a draft? As the draft would let us actually fill in the body, sign it, etc.

tanx:
@valentinewallace bryan just told me that cross compiling lnd for different platforms is actually super simple. See the release shell script for lnd: https://github.com/lightningnetwork/lnd/blob/master/release.sh#L24

valentinewallace:
@tanx wow, look good! def thought it would be trickier than that. it looks like that compiles for a ton of different platforms, i'm assuming we want that rather than just 1 build per linux/win/darwin like before? 

@Roasbeef yes that is possible! i'll switch it to that

tanx:
> it looks like that compiles for multiple specific linux x64 platforms, i'm assuming we want that rather than just 1 build per linux/win/darwin like before?

Yes. the app build targets are only linux/win/darwin. So: `SYS="windows-386 linux-amd64 darwin-amd64"`

.travis.yml
```diff
@@ -1,15 +1,57 @@
...
+  - env | grep -iE 'DEBUG|NODE_|ELECTRON_|YARN_|NPM_|CI|CIRCLE|TRAVIS_TAG|TRAVIS|TRAVIS_REPO_|TRAVIS_BUILD_|TRAVIS_BRANCH|TRAVIS_PULL_REQUEST_|APPVEYOR_|CSC_|GH_|GITHUB_|BT_|AWS_|STRIP|BUILD_' > env.txt
+  - docker run --rm --env-file env.txt --env ELECTRON_CACHE="/root/.cache/electron" --env ELECTRON_BUILDER_CACHE="/root/.cache/electron-builder" -v ${PWD}:/project -v ${PWD##*/}-node-modules:/project/node_modules -v ~/.cache/electron:/root/.cache/electron -v ~/.cache/electron-builder:/root/.cache/electron-builder electronuserland/builder:wine /bin/bash -c "npm i; npm run build; react-scripts build; ./node_modules/.bin/electron-builder --em.main=build/electron.js --linux tar.gz --win zip --mac zip"
+  - rm env.txt
+  - PACKAGE_VERSION=$(node -pe "require('./package.json').version")
+  - cd dist
+  - shasum -a 256 Lightning* | sudo tee manifest-v${PACKAGE_VERSION}.txt
+  - sudo chown -R travis:travis ./
+  - cd ..
+  - ls -R assets
+  - cat .gitignore
```
tanx: Would you mind moving this stuff to `./assets/script/build_lnd.sh`?

.travis.yml
```diff
@@ -1,15 +1,57 @@
...
+  provider: releases
+  api_key:
+    secure: Wy54wEpTwDWDPHCXtG0TOqokurh7oCvjxXgbkMumc98Mhpcj6GfkDcsFAD/6d4byUwoNFddWKLEwvrJlvQ2mFdWyi8PVaVtII5UA+mhQlqubsLG8YbiNXv0AhDnmibEhj8clQiIE8MQ6dHVXgLNrV+rbOUoS0hlArhduTdW1gDQQ9POxvyjtUQWINGzJGM3O4+976pLsV1476H7ShVect6fEoBkypPpYWiKeJUb44/KJT7QFEfkLs908z+iVeGlKTrYj+v8RnCpTGPKi0Dh8BzTrMx3wKB/xig7f8Jrh5gWQAM3PWQ2WHjeel7eXumTWP7k0cxolJbQ1o87b3vnmM9H6Dv2tlf9H6ojg5rMEXARgXNlqMKCcaMPMm/DFCnKkQbKGF6slflgzvM+oTcPVKBEgIfAB0407YWbZ88v4TkglMgT9cHB3cm9i9724ExVbR14HtcdUaTkQpb++/9TPsSdNt2Mttp3GQTKea0hmDXt4RAjieLkseks9Tqw08c6SVV7Uazbd02nds8u7HKYgy+juJ9z7kgFepW+/INdcmsPfWykKoBtKazBYwQ+RC3X1kE1Agbg9WSMh1zkuKn/38ZN9bik6nyrWB10gB5eVSiidtMiM433y6r0dgvT16lP0e1UTM/4e7eChvtvGjOBkvTqvD39uIxy/Tyq0iP+8sEk=
+  file_glob: true
+  file:
+    - dist/Lightning*
+    - dist/manifest*
+  skip_cleanup: true
+  on:
+    repo: valentinewallace/lightning-app
```
tanx: --> `lightninglabs/lightning-app`

valentinewallace:
Thank you, sorry that was meant to be a work-in-progress push. Will incorporate & squash & etc 👌

valentinewallace:
I had to pull in some newer changes to get lnd to build due to the switch from `glide` to `dep`, so rewrote the commit history pretty significantly, hopefully ok!

PR #291, opened 2018-04-03
Create Payment Request view

valentinewallace:
Also, modify BalanceLabel and IconButton to take in more style parameters in order to make them reusable for this and future views (while maintaining backwards compatibility).


src/view/payment-request.js
```diff
@@ -0,0 +1,80 @@
...
+import { BackButton, CopyButton, TextButton } from '../component/button';
+import { BalanceLabel } from '../component/label';
+import Card from '../component/card';
+import QRCode from '../component/qrcode';
+import { colors } from '../component/style';
+
+const styles = StyleSheet.create({
+  content: {
+    alignItems: 'stretch',
+    width: '98%',
```
tanx: Same comment about `%` as above...

src/view/payment-request.js
```diff
@@ -0,0 +1,80 @@
...
+  <Background image="purple-gradient-bg">
+    <Header>
+      <BackButton onPress={() => {}} />
+      <Title title="Payment Request" />
+      <View style={styles.hiddenItem}>
+        <BackButton onPress={() => {}} />
+      </View>
+    </Header>
+    <Card>
+      <MainContent style={styles.content}>
```
tanx: Any reason we need a `MainContent` component inside `Card`?

src/view/payment-request.js
```diff
@@ -0,0 +1,80 @@
...
+  </Background>
+);
+
+PaymentRequest.propTypes = {
+  amount: PropTypes.string,
+  message: PropTypes.string,
+  invoice: PropTypes.string,
+};
+
+export default PaymentRequest;
```
tanx: You need to wrap the constructor in a mobx `observer` call here. Otherwise the view won't re-render when the state changes. See: https://github.com/lightninglabs/lightning-app/blob/0e40f90e02bfae48ec6502dab40765cf65defc66/src/view/home.js#L115

stories/screen.js
```diff
@@ -3,12 +3,20 @@ import { storiesOf } from '@storybook/react';
...
 
 storiesOf('Screens', module)
   .add('Welcome', () => <WelcomeView />)
-  .add('Home', () => <HomeView store={store} />);
+  .add('Home', () => <HomeView store={store} />)
+  .add('Payment Request', () => (
+    <PaymentRequest
+      amount="456,780"
+      invoice="lnbc4567800n1pdvqx48pp5eng6uyqnkdlx93m2598ug93qtuls8gapygxznshzd56h7n5cxs0sdp9gehhygr5dpjjqmr0wejjqmmxyp3xjarrda5kucqzysmhyrleqpt3yqf5nctzsr3hvrv9vhhnawazkwyzu8t4mf85tllsyjsf8hgu5nt6dj3jaljjgmt999xnlsweqvatypzlu34nhpjlxf59qp4dn2pv"
+      message="For the love of bitcoin"
```
tanx: Injection individual attributes won't work unfortunately. You need to pass by reference not by value. Otherwise the mobx observer won't be notified of state updates. See: https://mobx.js.org/refguide/observer-component.html#gotcha-dereference-values-inside-your-components

Ideally we'd pass the whole mobx store into each view like so `<HomeView store={store} />`. That way we have a consistent way to render views from a single source of truth :)

tanx:
@valentinewallace I also tested it locally in react storybook. Comments below:
bildschirmfoto 2018-04-04 um 16 52 09
It seems like the header is missing the `shadow`.

bildschirmfoto 2018-04-04 um 16 52 14
This is what happens when I scroll down...

valentinewallace:
Hey @tanx, done with everything except for correcting the card size and header width on scroll. 
Question on 

> Expected behavior would be for the whole button to be clickable and the text to be selected after it's been clicked on.

In order to do this, I would need to make CopyButton stateful so it knows when it's clicked, so that i can tell the text to be highlighted. Is this acceptable? 

Asking because it goes against the conventions of the other code to make the component stateful. To me, an acceptable solution might be to leave the component stateless and make the text selected if the text itself is clicked, but if other parts of the button are clicked it won't be selected.

tanx:
> Hey @tanx, done with everything except for correcting the card size and header width on scroll.

@valentinewallace thank you for your changes! I haven't gotten around to reviewing everything yet. I just saw that there are a bunch or PropType Warnings in the console in storybook. If you could fix those that would be great. Also, it seems like there are merge conflicts again. Seems like we've been stepping on each others toes yesterday. Sorry about that. I cleaned up a few things in the components.

> In order to do this, I would need to make CopyButton stateful so it knows when it's clicked, so that i can tell the text to be highlighted. Is this acceptable?
> 
> Asking because it goes against the conventions of the other code to make the component stateful. To me, an acceptable solution might be to leave the component stateless and make the text selected if the text itself is clicked, but if other parts of the button are clicked it won't be selected.

No, that's alright. Looks like the whole button is now clickable. We can use the copy/paste apis once we wire the view up to the actions.

src/view/payment-request.js
```diff
@@ -0,0 +1,93 @@
...
+const styles = StyleSheet.create({
+  header: {
+    minWidth: 500,
+    width: '100%',
+    alignSelf: 'center',
+  },
+  content: {
+    alignItems: 'stretch',
+    justifyContent: 'flex-start',
+    overflow: 'auto',
```
tanx: This won't work in react-native unfortunately https://stackoverflow.com/questions/39722439/react-native-overflow-and-scroll?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

I've made `MainContent` scrollable via `ScrollView` in https://github.com/lightninglabs/lightning-app/commit/88354ab6345740fd25fda049a34eec2f77a392dd. Looks like we'll have to wrap our `Card` in `MainContent` after all. Here's an example of how to use it: https://github.com/lightninglabs/lightning-app/blob/5379750379741157be31d8a2cf53f3dbf53f5f21/stories/layout.js#L37-L49

src/component/style.js
```diff
@@ -11,6 +11,7 @@ export const colors = {
   blueSig: '#002FFF',
   glas: 'rgba(255,255,255,0.13)',
   glasDark: 'rgba(255,255,255,0.08)',
+  darkText: '#4A4A4A',
```
tanx: I've added `blackText` in https://github.com/lightninglabs/lightning-app/commit/4b33b5175eb9dd49ab2d765349be1167566dc9f1 to stay consistent with our naming.

src/view/payment-request.js
```diff
@@ -0,0 +1,93 @@
...
+import Card from '../component/card';
+import QRCode from '../component/qrcode';
+import { colors } from '../component/style';
+
+const styles = StyleSheet.create({
+  header: {
+    minWidth: 500,
+    width: '100%',
+    alignSelf: 'center',
+  },
```
tanx: You can actually remove the `header` stylesheet completely and it won't change anything :D

src/view/payment-request.js
```diff
@@ -0,0 +1,93 @@
...
+  content: {
+    alignItems: 'stretch',
+    justifyContent: 'flex-start',
+    overflow: 'auto',
+  },
+  amount: {
+    color: colors.darkText,
```
tanx: Please use the new `blackText` color. Thanks :)

src/view/payment-request.js
```diff
@@ -0,0 +1,93 @@
...
+    <Header shadow color={colors.purple} style={styles.header}>
+      <BackButton onPress={() => {}} />
+      <Title title="Payment Request" />
+      <View style={styles.hiddenItem}>
+        <BackButton onPress={() => {}} />
+      </View>
+    </Header>
+    <Card style={styles.content}>
+      <View style={styles.amountWrapper}>
+        <BalanceLabel unit={<Text style={styles.unit}>SAT</Text>}>
```
tanx: I haven't seen child nodes being injected via attributes. Maybe we can think of a cleaner way to do this. What are you trying to achieve?

tanx: Maybe we can use `adjustsfontsizetofit` somehow?
https://facebook.github.io/react-native/docs/text.html#adjustsfontsizetofit

stories/screen.js
```diff
@@ -6,16 +6,25 @@ import HomeView from '../src/view/home';
...
   .add('Welcome', () => <WelcomeView />)
-  .add('Home', () => <HomeView store={store} wallet={wallet} nav={nav} />);
+  .add('Home', () => <HomeView store={store} wallet={wallet} nav={nav} />)
+  .add('Payment Request', () => <PaymentRequest store={store} />);
 
 // set some dummy data
 store.balanceSatoshis = 798765432;
 store.channelBalanceSatoshis = 59876543;
 store.settings.exchangeRate.usd = 0.00014503;
+store.settings.displayFiat = true;
```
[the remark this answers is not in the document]
valentinewallace: Sorry bout it 💁‍♀️

src/view/payment-request.js
valentinewallace: I was trying to figure out how to style the unit only, but realized this was unnecessary. 

Looks like `adjustsFontSizeToFit` isn't supported in react-native-web: https://necolas.github.io/react-native-web/storybook/?selectedKind=Components&selectedStory=Text&full=0&addons=0&stories=1&panelRight=0

One potential problem that remains is that if someone's balance is greater than 999,999 satoshis then the full balance won't fit. I think an acceptable solution could be either to use a smaller font size or to calculate the appropriate font size based on screen widths?

src/component/label.js
```diff
@@ -34,7 +34,7 @@ export const BalanceLabel = ({ children, unit, style }) => (
 );
 
 BalanceLabel.propTypes = {
-  children: PropTypes.string.isRequired,
+  children: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
```
tanx: Shouldn't be necessary anymore with new `BalanceLabel`.

src/view/payment-request.js
```diff
@@ -0,0 +1,88 @@
...
+      <BackButton onPress={() => {}} />
+      <Title title="Payment Request" />
+      <Button disabled onPress={() => {}} />
+    </Header>
+    <MainContent>
+      <Card style={styles.card}>
+        <View style={styles.amountWrapper}>
+          <BalanceLabel unit="SAT" style={styles.unit}>
+            <Text style={styles.numeral}>{store.paymentRequest.amount}</Text>
+          </BalanceLabel>
```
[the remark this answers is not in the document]
valentinewallace: This is nice :+1:

tanx:
@valentinewallace great work! And thanks for hanging in there. It took a few iterations but the code quality is really solid now and I also got some valuable feedback on my components. Should make things much more reusable for building other screens. 👍

valentinewallace:
love eet!!!

PR #305, opened 2018-04-21
Get current exchange rate

### LINKED ISSUE #290: Get current exchange rates for BTC via an API
e.g. https://blockchain.info/tobtc?currency=USD&value=1

tanx:
Closes #290

@halseth this uses the `blockchain.info` api. Let me know if you have any thoughts


tanx:
> utACK as initial solution.

Ok. Merging for now.

> Maybe long term we could consider fetching rate from multiple sources, to not rely on one centralized one?

Yeah I've looked at https://bitcoinaverage.com/en/plans but they're not free.

PR #351, opened 2018-05-30
implement Loader - First Time screen

### LINKED ISSUE #250: Implement Loader - First Time Screen
https://projects.invisionapp.com/d/main/default/#/console/13241316/281222443/preview

valentinewallace:
Closes #250

### DIFF
--- a/src/view/loader.js
+++ b/src/view/loader.js
@@ -1,0 +1,53 @@
+import React from 'react';
+import { StyleSheet, View } from 'react-native';
+import Background from '../component/background';
+import { CopyOnboardText } from '../component/text';
+import MainContent from '../component/main-content';
+import Icon from '../component/icon';
+import { color } from '../component/style';
+
+const styles = StyleSheet.create({
+  content: {
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+  copy: {
+    flexDirection: 'row',
+    justifyContent: 'center',
+    alignItems: 'center',
+    width: 900,
+    paddingLeft: 30,
+    paddingRight: 30,
+  },
+  boltBackground: {
+    alignItems: 'center',
+    justifyContent: 'center',
+    height: 46,
+    width: 25,
+    marginLeft: 26,
+    backgroundColor: color.purple,
+  },
+  bolt: {
+    height: 37,
+    width: 19,
+  },
+});
+
+const LoaderView = () => (
+  <Background color={color.blackDark}>
+    <MainContent style={styles.content}>
+      <View style={styles.copy}>
+        <CopyOnboardText>The fastest way to transfer Bitcoin.</CopyOnboardText>
+        <Bolt />
+      </View>
+    </MainContent>
+  </Background>
+);
+
+const Bolt = () => (
+  <View style={styles.boltBackground}>
+    <Icon image="lightning-bolt" style={styles.bolt} />
+  </View>
+);
+
+export default LoaderView;
--- a/src/component/text.js
+++ b/src/component/text.js
@@ -51,6 +51,27 @@ CopyText.propTypes = {
 };
 
 //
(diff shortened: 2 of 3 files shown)


tanx:
This is how it renders. I'll make comments inline...

image

src/view/loader.js
```diff
@@ -0,0 +1,44 @@
...
+  },
+  bolt: {
+    height: 37,
+    width: 19,
+  },
+});
+
+const LoaderView = () => (
+  <Background color={color.blackDark}>
+    <FormStretcher>
```
tanx: Not sure `FormStretcher` is the sematics we want here. It's meant to be used for forms. If we refactor/optimize it for form use, we'll end up changing this screen, which would not be ideal.

In any case, we should always wrap views content in a scrollable `MainContent` in case the content height is bigger than the viewport.

src/view/loader.js
```diff
@@ -0,0 +1,44 @@
...
+      <CopyText style={styles.copyTxt}>
+        The fastest way to transfer Bitcoin.<Bolt />
+      </CopyText>
```
tanx: We should add some padding left and right of the text. This is probably best achieved with an extra container of a fixed width that is centered by the `MainContent` parent component.

src/view/loader.js
```diff
@@ -0,0 +1,44 @@
...
+import Background from '../component/background';
+import { CopyText } from '../component/text';
+import { FormStretcher } from '../component/form';
+import Icon from '../component/icon';
+import { color } from '../component/style';
+
+const styles = StyleSheet.create({
+  copyTxt: {
+    flexDirection: 'row',
+    fontSize: 53,
```
tanx: We should try to avoid setting hard coded font sizes in the views directly and use predefined font sizes from:
https://github.com/lightninglabs/lightning-app/blob/master/src/component/style.js#L25

src/view/loader.js
```diff
@@ -0,0 +1,44 @@
...
+import { StyleSheet, View } from 'react-native';
+import Background from '../component/background';
+import { CopyText } from '../component/text';
+import { FormStretcher } from '../component/form';
+import Icon from '../component/icon';
+import { color } from '../component/style';
+
+const styles = StyleSheet.create({
+  copyTxt: {
+    flexDirection: 'row',
```
tanx: A text component should never be used as a wrapper for non-text components. Instead add a wrapper around `Bolt` and `CopyText`.

src/view/loader.js
```diff
@@ -0,0 +1,44 @@
...
+  bolt: {
+    height: 37,
+    width: 19,
+  },
+});
+
+const LoaderView = () => (
+  <Background color={color.blackDark}>
+    <FormStretcher>
+      <CopyText style={styles.copyTxt}>
```
tanx: CopyText has the wrong font. We need `WorkSans Light` here. Maybe create a new `CopyOnboard` component in text similar to H1Text:
https://github.com/lightninglabs/lightning-app/blob/master/src/component/text.js#L53

src/view/loader.js
```diff
@@ -0,0 +1,44 @@
...
+const styles = StyleSheet.create({
+  copyTxt: {
+    flexDirection: 'row',
+    fontSize: 53,
+  },
+  boltBackground: {
+    height: 46,
+    width: 25,
+    marginLeft: 26,
+    backgroundColor: '#57038D',
```
tanx: Please always reuse/define colors in: https://github.com/lightninglabs/lightning-app/blob/master/src/component/style.js#L1

src/view/loader.js
valentinewallace: Hopefully what i went with is what you meant by that, i made a fixed width container. lmk if not!

PR #352, opened 2018-05-30
Close #251: Implement seed screen

### LINKED ISSUE #251: Implement Seed Screen
https://projects.invisionapp.com/d/main/default/#/console/13241316/281903496/preview

valentinewallace:
Hey, I was spending way too long figuring out how to get the SplitBackground to scale with the height of the content so I made a compromise and made the bottom half scrollable. LMK if you have any tips on that :)


src/component/button.js
```diff
@@ -78,7 +78,13 @@ const glasStyles = StyleSheet.create({
   },
 });
 
-export const GlasButton = ({ onPress, disabled, children, style }) => (
+export const GlasButton = ({
+  onPress,
+  disabled,
+  children,
+  style,
+  textStyle,
```
tanx: We should avoid setting two style objects in one component. If necessary, I'd prefer pulling the inner components out to make them styleable. But it actually shouldn't be necessary to style the text size for this button.

tanx: In general the fonts are pretty big. We should try to fit the design into the default viewport size https://github.com/lightninglabs/lightning-app/blob/dd046e284bf13b8c09653eca5c29f5bcb8039d05/public/electron.js#L73-L74

This is what that currently looks like...
image

src/view/seed.js
```diff
@@ -0,0 +1,111 @@
...
+  title: {
+    padding: 20,
+    fontSize: font.sizeXXL + 2 * 6,
+    textAlign: 'center',
+  },
+  copyTxt: {
+    alignSelf: 'center',
+    textAlign: 'center',
+    fontSize: font.sizeL + 2 * 2,
+    lineHeight: font.lineHeightL + 2 * 4,
```
tanx: If you need to multiply the fonts by a constant factor, that's probably a good sign that we need a new font size. Considering how large the seed view renders on the default viewport size though, I'm pretty sure we can do with the font sizes we currently already have.

src/component/list.js
```diff
@@ -55,6 +55,9 @@ export class List extends Component {
         renderHeader={this.props.renderHeader}
         renderRow={this.props.renderItem}
         enableEmptySections={true}
+        horizontal={this.props.horizontal}
```
tanx: Why do we want the list to scroll horizontally?

valentinewallace: True, I forget why i put that in. took out

tanx:
I added some minor changes in the two commits.

tanx:
I added some minor changes (see my commits).

> Hey, I was spending way too long figuring out how to get the SplitBackground to scale with the height of the content so I made a compromise and made the bottom half scrollable. LMK if you have any tips on that :)

I changed it so the the top container is a flex View instead of having two MainContent components in one screen, which doesn't really make much sense semantically.

src/view/seed.js
```diff
@@ -0,0 +1,99 @@
...
+    paddingBottom: 20,
+    maxWidth: 600,
+  },
+  wordList: {
+    flexDirection: 'row',
+    flexWrap: 'wrap',
+    justifyContent: 'center',
+    alignItems: 'center',
+    paddingTop: 25,
+    paddingBottom: 25,
```
tanx: I'm wondering if it's possible to have the List float in the center of the MainView instead of adjusting the position via paddingTop?

stories/screen.js
```diff
@@ -102,7 +103,8 @@ storiesOf('Screens', module)
   .add('Invoice QR', () => (
     <InvoiceQR store={store} invoice={invoice} nav={nav} />
   ))
-  .add('Loader - First Time', () => <Loader />);
+  .add('Loader - First Time', () => <Loader />)
+  .add('Seed', () => <Seed seedMnemonic={store.seedMnemonic} />);
```
tanx: We should stay consistent and always inject the complete here.

src/view/seed.js
```diff
@@ -0,0 +1,99 @@
...
+    </View>
+    <MainContent>
+      <List
+        data={seedMnemonic.slice()}
+        renderItem={(word, _, rowID) => (
+          <Word word={word} ind={parseInt(rowID) + 1} />
+        )}
+        contentStyle={styles.wordList}
+        initialListSize={24}
+      />
```
tanx: I'm also wondering whether a ListView is the right component here. Seems like it would be easier to just render the 24 boxes in a forEach loop inside a View. Then we wouldn't have to style the ListView either.

src/view/seed.js
```diff
@@ -0,0 +1,124 @@
...
+//
+
+const SeedView = ({ store }) => (
+  <SplitBackground image="purple-gradient-bg" bottom={color.blackDark}>
+    <CopySection />
+    <MainContent>
+      <WordList seedMnemonic={store.seedMnemonic.slice()} />
+      <GlasButton onPress={() => {}}>Next</GlasButton>
+    </MainContent>
+  </SplitBackground>
```
tanx: I split up the view into components.

src/view/seed.js
```diff
@@ -0,0 +1,124 @@
...
+    justifyContent: 'center',
+    alignItems: 'center',
+  },
+  title: {
+    marginBottom: 15,
+  },
+  copyTxt: {
+    textAlign: 'center',
+    maxWidth: 410,
+  },
```
tanx: I also cleaned up the font sizes to be in line with the rest of the instances.

src/view/seed.js
```diff
@@ -0,0 +1,124 @@
...
+});
+
+const WordList = ({ seedMnemonic }) => (
+  <View style={listStyles.wrapper}>
+    <View style={listStyles.words}>
+      {seedMnemonic.map((word, i) => (
+        <Word word={word} index={i + 1} key={i} />
+      ))}
+    </View>
+  </View>
```
tanx: This wasn't rendering correctly. So I changed the styled. Also `key` is a reserved attribute and can't be used as a prop...

image

src/view/seed.js
valentinewallace: Ahh, good one

src/view/seed.js
valentinewallace: Hmm, that's weird, i was getting a console error saying `key` was required when you're using a `for` loop like that... i'll re-run it

valentinewallace:
Great feedback, thanks. I rebased onto master and added a minor `title` style change to prevent this edge case: 
first_write_down2

tanx:
> Great feedback, thanks. I rebased onto master and added a minor title style change to prevent this edge case:

@valentinewallace good catch. I think this is what you meant https://github.com/lightninglabs/lightning-app/commit/7d2c55c97619cc1a6454f2d643b84a5491a1f8a1 ?  If you have changes/comments we can keep iterating on the branch next time until it's ready to merge.

Other than that great to finally get this merged. Good job 👍

PR #358, opened 2018-06-01
Update deprecated flag and homepage field

valentinewallace:
I copied the changes over from my Ubuntu VM so hopefully they work. 

Additionally, I had to `cp` my lnd and btcd binaries into the `assets/build/linux` folder for the package to build correctly (symlinks didn't work). Not sure if this should be put into the README or what?


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "lightning-app",
   "version": "0.1.0",
   "description": "Lightning Wallet Application",
-  "homepage": "https://lightning.engineering",
+  "homepage": "./",
```
tanx: The npm docs say this attribute is for the project homepage? https://docs.npmjs.com/files/package.json#homepage

package.json
```diff
@@ -16,7 +16,7 @@
     "test:react": "react-scripts test --env=jsdom",
     "eject": "react-scripts eject",
     "electron-dev": "concurrently \"BROWSER=none npm start\" \"wait-on http://localhost:3000 && electron --enable-sandbox .\"",
-    "electron-pack": "build --em.main=build/electron.js",
+    "electron-pack": "build --c.extraMetadata.main=build/electron.js",
```
tanx: So this fixes the build?

tanx:
> Additionally, I had to cp my lnd and btcd binaries into the assets/build/linux folder for the package to build correctly (symlinks didn't work). Not sure if this should be put into the README or what?

Do you mean during development or for the build? During development you just have to make sure you have lnd/btcd in your `PATH`. This is explained in the installation guide which is already linked in the README: https://github.com/lightninglabs/lightning-app/blob/master/README.md#developing-locally

package.json
valentinewallace: Both changes are part of the fix. However, the line this change replaces is the reason Travis isn't uploading any files on releases

package.json
valentinewallace: [https://github.com/facebook/create-react-app/blob/ca7d227ae0354e04941ba0e4f76c0bcbf3050dfa/config/webpack.config.prod.js#L20-L23](url) create-react-app uses `package.json`'s `homepage` field when it creates the webpack config as webpack's `publicPath`, which is what the static files are served from. Got this fix from here: https://github.com/facebook/create-react-app/issues/165 lmk if it isn't acceptable

valentinewallace:
> Do you mean during development or for the build? During development you just have to make sure you have lnd/btcd in your PATH. This is explained in the installation guide which is already linked in the README: https://github.com/lightninglabs/lightning-app/blob/master/README.md#developing-locally

I mean for the build of the packaged app, but in both cases i needed to do this. i'm confused why it would work otherwise due to the structure of `getProcessName` in `lnd-child-process` -- messaging you

tanx:
@valentinewallace Ok. Tested this. Looks good 👍

PR #365, opened 2018-06-05
Closes #252: Implement seed verify screen

### LINKED ISSUE #252: Implement Seed Verify Screen
https://projects.invisionapp.com/d/main/default/#/console/13241316/281903495/preview

valentinewallace:
Additionally, compute indices of seed to check and add a test for this. Also added more style props to `GlasButton` and `FormStretcher`

### DIFF
--- a/src/component/form.js
+++ b/src/component/form.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { View, TextPropTypes, StyleSheet } from 'react-native';
+import { View, TextPropTypes, ViewPropTypes, StyleSheet } from 'react-native';
 import PropTypes from 'prop-types';
 import Text from './text';
 import { color, font } from './style';
@@ -19,12 +19,13 @@ const stretcherStyles = StyleSheet.create({
   },
 });
 
-export const FormStretcher = ({ children }) => (
-  <View style={stretcherStyles.content}>{children}</View>
+export const FormStretcher = ({ children, style }) => (
+  <View style={(stretcherStyles.content, style)}>{children}</View>
 );
 
 FormStretcher.propTypes = {
   children: PropTypes.node,
+  style: ViewPropTypes.style,
 };
 
 //
--- a/src/component/form.js
+++ b/src/component/form.js
@@ -59,8 +59,6 @@ const subTextStyles = StyleSheet.create({
     lineHeight: font.lineHeightSub,
     color: color.greyText,
     textAlign: 'center',
-    paddingTop: 40,
-    paddingBottom: 40,
   },
 });
 
--- a/src/component/form.js
+++ b/src/component/form.js
@@ -20,7 +20,7 @@ const stretcherStyles = StyleSheet.create({
 });
 
 export const FormStretcher = ({ children, style }) => (
-  <View style={(stretcherStyles.content, style)}>{children}</View>
+  <View style={[stretcherStyles.content, style]}>{children}</View>
 );
 
 FormStretcher.propTypes = {
--- a/src/computed/seed.js
+++ b/src/computed/seed.js
@@ -2,13 +2,27 @@ import { computed, extendObservable } from 'mobx';
 
 const ComputedSeed = store => {
   extendObservable(store, {
-    seedCheck: computed(() => {
+    seedVerifyIndexes: computed(() => {
       const { seedMnemonic: words } = store;
       return words.length ? getSeedCheck(1, words.length, 3) : [];
+    }),
+    seedVerifyCopy: computed(() => {
+      const { seedVerifyIndexes } = store;
+      const c0 = formatOrdinal(seedVerifyIndexes[0]);
+      const c1 = formatOrdinal(seedVerifyIndexes[1]);
+      const c2 = formatOrdinal(seedVerifyIndexes[2]);
+      return `Type the ${c0}, ${c1}, and ${c2} words of your seed.`;
     }),
   });
 };
 
+/**
+ * Get an array of seed word indexes to verify
+ * @param  {number} min      The smallest possible index
+ * @param  {number} max      The largest possible index
+ * @param  {number} numWords The length of the array to generate
+ * @return {Array}           The array of indexes to verify
+ */
 const getSeedCheck = (min, max, numWords) => {
   let indices = [];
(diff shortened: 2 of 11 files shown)


src/computed/seed.js
```diff
@@ -0,0 +1,27 @@
+import { computed, extendObservable } from 'mobx';
+
+const ComputedSeed = store => {
+  extendObservable(store, {
+    seedCheck: computed(() => {
+      const { seedMnemonic: words } = store;
+      return words ? getSeedCheck(1, words.length, 3) : null;
```
tanx: Probably better return an empty array here.

src/helper.js
```diff
@@ -27,6 +27,25 @@ export const formatFiat = (val, currency) => {
...
+  }
+  if (j == 2 && k != 12) {
+    return val.toString() + 'nd';
+  }
+  if (j == 3 && k != 13) {
+    return val.toString() + 'rd';
+  }
+  return val.toString() + 'th';
+};
+
```
tanx: Unit tests? :)

src/view/seed-verify.js
```diff
@@ -0,0 +1,102 @@
...
+    <Text style={styles.leaderTxt}>{leaderText}.</Text>
+    <InputField style={styles.input} onChangeText={() => {}} />
+  </FormStretcher>
+);
+
+SeedEntry.propTypes = {
+  leaderText: PropTypes.string,
+};
+
+export default observer(SeedVerifyView);
```
tanx: I'm getting a bunch of errors in the console...

image

stories/screen.js
```diff
@@ -102,7 +103,8 @@ storiesOf('Screens', module)
   .add('Invoice QR', () => (
     <InvoiceQR store={store} invoice={invoice} nav={nav} />
   ))
-  .add('Loader - First Time', () => <Loader />);
+  .add('Loader - First Time', () => <Loader />)
+  .add('Seed Verify', () => <SeedVerify seedCheck={store.seedCheck} />);
```
tanx: This won't work if `store.seedCheck` changes, as it will reference an old instance. We should inject the `store` here to stay consistent with other views.

stories/screen.js
```diff
@@ -165,3 +167,29 @@ store.pendingChannels = [...Array(6)].map((x, i) => ({
...
+  'manager',
+  'tool',
+  'fan',
+  'neglect',
+  'conduct',
+  'blouse',
+  'stone',
+  'quit',
+  'cashew',
+];
```
tanx: Please rebase on top of the master after merging the `SeedView` since this is probably redundant in this branch. In general if once branch builds on top of another, you can just rebase them on top of each other and we'll review/merge them in order.

valentinewallace: Ah that's a good idea :+1:

src/helper.js
valentinewallace: Hehe yes

stories/screen.js
valentinewallace: my bad, forgot to make this change consistently when going through my PR backlog :sweat_smile:

tanx:
@valentinewallace I made a few changes. Please take a look. I haven't finished my review yet though. Still need to check out the logic behind the markup.

src/view/seed-verify.js
```diff
@@ -0,0 +1,98 @@
...
+    <Header>
+      <BackButton onPress={() => {}} />
+      <Button disabled onPress={() => {}} />
+    </Header>
+    <MainContent style={styles.content}>
+      <View>
+        <H1Text style={styles.title}>{"Let's double check"}</H1Text>
+      </View>
+      <Card style={styles.card}>
+        <FormSubText>{store.seedVerifyCopy}</FormSubText>
```
tanx: I refactor the ComputedSeed to generate the complete copy and pulled the logic out of the View. This way we keep the React component as dumb as possible.

src/computed/seed.js
```diff
@@ -0,0 +1,65 @@
...
+    return num.toString() + 'st';
+  }
+  if (j == 2 && k != 12) {
+    return num.toString() + 'nd';
+  }
+  if (j == 3 && k != 13) {
+    return num.toString() + 'rd';
+  }
+  return num.toString() + 'th';
+};
```
tanx: I moved this guy here, since it probably isn't going to be used anywhere else.

src/computed/seed.js
```diff
@@ -0,0 +1,65 @@
...
+  });
+};
+
+/**
+ * Get an array of seed word indexes to verify
+ * @param  {number} min      The smallest possible index
+ * @param  {number} max      The largest possible index
+ * @param  {number} numWords The length of the array to generate
+ * @return {Array}           The array of indexes to verify
+ */
```
tanx: I added some JSdocs here. We should probably do this for all new code being written. I've been slacking on that front up until now.

src/computed/seed.js
```diff
@@ -0,0 +1,65 @@
...
+      const { seedMnemonic: words } = store;
+      return words.length ? getSeedIndexes(1, words.length, 3) : [];
+    }),
+    seedVerifyCopy: computed(() => {
+      const { seedVerifyIndexes } = store;
+      const c0 = formatOrdinal(seedVerifyIndexes[0]);
+      const c1 = formatOrdinal(seedVerifyIndexes[1]);
+      const c2 = formatOrdinal(seedVerifyIndexes[2]);
+      return `Type the ${c0}, ${c1}, and ${c2} words of your seed.`;
+    }),
```
tanx: Renamed some vars here to be more descriptive.

src/component/form.js
```diff
@@ -58,8 +59,6 @@ const subTextStyles = StyleSheet.create({
     lineHeight: font.lineHeightSub,
     color: color.greyText,
     textAlign: 'center',
-    paddingTop: 40,
-    paddingBottom: 40,
```
tanx: I removed padding in from `FormSubText` so that we can use it more generically e.g. in our SeedVerifyView.

src/view/seed-verify.js
valentinewallace: good point

PR #369, opened 2018-06-06
Settings View

### LINKED ISSUE #265: Implement Settings Screen
https://projects.invisionapp.com/d/main/default/#/console/13241316/281222455/preview

tanx:
Closes #265

### DIFF
--- a/src/component/notification.js
+++ b/src/component/notification.js
@@ -97,9 +97,10 @@ const countStyles = StyleSheet.create({
   },
 });
 
-export const CountBubble = ({ children, style }) => (
-  <H4Text style={[countStyles.bubble, style]}>{children}</H4Text>
-);
+export const CountBubble = ({ children, style }) =>
+  children && children !== '0' ? (
+    <H4Text style={[countStyles.bubble, style]}>{children}</H4Text>
+  ) : null;
 
 CountBubble.propTypes = {
   children: PropTypes.string.isRequired,
--- a/src/component/notification.js
+++ b/src/component/notification.js
@@ -81,6 +81,32 @@ Alert.propTypes = {
 };
 
 //
+// CountBubble
+//
+
+const countStyles = StyleSheet.create({
+  bubble: {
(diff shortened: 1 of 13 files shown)


src/component/notification.js
```diff
@@ -80,6 +80,33 @@ Alert.propTypes = {
...
+    paddingBottom: 2,
+    paddingRight: 10,
+    paddingLeft: 10,
+    borderRadius: 13,
+    backgroundColor: color.pinkSig,
+  },
+});
+
+export const CountBubble = ({ children, style }) =>
+  children && children !== '0' ? (
```
valentinewallace: In the past I've benefited from setting a `display: none` or `opacity: 0` property in these situations, to prevent the missing component from unexpectedly rearranging other components. but might not matter here

stories/screen.js
```diff
@@ -112,6 +114,11 @@ store.invoice.note = 'For the love of bitcoin';
 store.invoice.encoded =
   'lnbc4567800n1pdvqx48pp5eng6uyqnkdlx93m2598ug93qtuls8gapygxznshzd56h7n5cxs0sdp9gehhygr5dpjjqmr0wejjqmmxyp3xjarrda5kucqzysmhyrleqpt3yqf5nctzsr3hvrv9vhhnawazkwyzu8t4mf85tllsyjsf8hgu5nt6dj3jaljjgmt999xnlsweqvatypzlu34nhpjlxf59qp4dn2pv';
 store.invoice.uri = `lightning:${store.invoice.encoded}`;
+store.notifications = [...Array(5)].map((x, i) => ({
+  type: i % 2 === 0 ? 'error' : 'info',
+  message: 'Ooops. Something went wrong.',
```
valentinewallace: Doesn't really matter, but... "Ooops" --> "Oops"

src/component/notification.js
tanx: That's a good point for elements that relevant for layout. In this case the count bubble should actually only render for non zero values.

PR #377, opened 2018-06-08
Implement small spinner

valentinewallace:
@tanx what do you think about using these for our purple spinners? they can't do the gradient from the invision project but they should be able to scale to the correct size

### DIFF
--- a/src/component/spinner.js
+++ b/src/component/spinner.js
@@ -1,0 +1,11 @@
+import React from 'react';
+import { ActivityIndicator } from 'react-native';
+import { color } from './style';
+
+//
+// Small Spinner
+//
+
+export const SmallSpinner = () => (
+  <ActivityIndicator size="large" color={color.purpleSpinner} />
+);
--- a/stories/component/spinner.js
+++ b/stories/component/spinner.js
@@ -1,0 +1,10 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import MainContent from '../../src/component/main-content';
+import { SmallSpinner } from '../../src/component/spinner';
+
+storiesOf('Spinner', module)
+  .addDecorator(story => (
+    <MainContent style={{ justifyContent: 'center' }}>{story()}</MainContent>
+  ))
+  .add('SmallSpinner', () => <SmallSpinner />);
--- a/src/component/style.js
+++ b/src/component/style.js
@@ -22,6 +22,7 @@ export const color = {
   glasDarker: 'rgba(255,255,255,0.04)',
   notifyDark: 'rgba(37,47,74,0.5)',
   notifyLight: 'rgba(255,255,255,0.5)',
+  purpleSpinner: '#651399',
 };
 
 export const font = {


tanx:
> @tanx what do you think about using these for our purple spinners? they can't do the gradient from the invision project but they should be able to scale to the correct size

Mmmh. Probably good enough for a first iteration. Let me test it out.

stories/component/spinner.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import MainContent from '../../src/component/main-content';
+import { SmallSpinner } from '../../src/component/spinner';
+
+storiesOf('Spinner', module)
+  .addDecorator(story => (
+    <MainContent style={{ justifyContent: 'center' }}>{story()}</MainContent>
+  ))
+  .add('SmallSpinner', () => <SmallSpinner />);
```
tanx: I tried to set height/width via the `style` attribute but I couldn't scale it 🤔

valentinewallace:
Added scaling :)

I'm working on making the percentage-based medium spinners using this tutorial: https://cmichel.io/react-native-progress-circle

valentinewallace:
Hey, using the ART library I couldn't get the linear gradient to work. are you opposed to using this library? https://github.com/godaddy/svgs

src/component/spinner.js
```diff
@@ -0,0 +1,87 @@
...
+  p.move(cx + r, cy);
+  p.arc(
+    -r + r * Math.cos(endDegreeInRadians),
+    r * Math.sin(endDegreeInRadians),
+    r,
+    r,
+    endDegree >= 180
+  );
+  return p;
+};
```
tanx: Nice. Could we begin the circle from to top/center like in the design?
image

src/component/spinner.js
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+import {
+  ActivityIndicator,
+  StyleSheet,
+  View,
+  ART,
+} from 'react-native';
```
tanx: Seems like there are currently linter errors in the build https://travis-ci.org/lightninglabs/lightning-app/jobs/394053652#L4509

valentinewallace:
@tanx got the gradient working! still working on how to modularize the components to make them reusable for the remaining spinners

src/component/spinner.js
```diff
@@ -0,0 +1,126 @@
+import React from 'react';
+import { ActivityIndicator, StyleSheet, View } from 'react-native';
+import PropTypes from 'prop-types';
+import { color, font } from './style';
+import Icon from '../component/icon';
+import Text from '../component/text';
+import Svg, { Path, Circle, Defs, Stop, LinearGradient } from 'svgs';
```
tanx: Given that `svgs` is web only, I’d wrap it and re-export it in an `svg.js` component. That way we can implement a `svg.ios.js` and `svg.android.js` components that re-export `react-native-svg`. See https://facebook.github.io/react-native/docs/platform-specific-code.html

package.json
```diff
@@ -37,7 +37,8 @@
     "react-art": "^16.2.0",
     "react-dom": "^16.2.0",
     "react-native-web": "^0.6.0",
-    "react-scripts": "^1.1.1"
+    "react-scripts": "^1.1.1",
+    "svgs": "^3.2.1"
```
tanx: Seems like the package-lock.json shouldn’t change so much here. Which node/npm version are you using? We should stick with the current LTS node@8.x/npm@5.x since newer versions of npm will rebuild the package-lock.json and potentially cause incompapatibilites on travis.

src/component/spinner.js
```diff
@@ -0,0 +1,133 @@
...
+import { color, font } from './style';
+import Icon from '../component/icon';
+import Text from '../component/text';
+import Svg, {
+  Path,
+  Circle,
+  Defs,
+  Stop,
+  LinearGradient,
+} from '../component/svg';
```
tanx: ./svg from the components dir (for text/icon too)

src/component/svg.js
```diff
@@ -0,0 +1,16 @@
...
+} from 'svgs';
+
+export const Path = PathJS;
+export const Circle = CircleJS;
+export const Defs = DefsJS;
+export const Stop = StopJS;
+export const LinearGradient = LinearGradientJS;
+
+const Svg = SvgJS;
+export default Svg;
```
tanx: We should re-export all here. Try

export * from 'svgs';

See https://stackoverflow.com/questions/29844074/es6-export-all-values-from-object

package-lock.json
```diff
@@ -14724,6 +14729,15 @@
...
     },
+    "svgs": {
+      "version": "3.2.1",
+      "resolved": "https://registry.npmjs.org/svgs/-/svgs-3.2.1.tgz",
+      "integrity": "sha512-+bsyTsin64hdOpxNOE8kClBW1GUujkSgoFcbEJz/dDa2bgKlyloPVqolTyPw+QPEIQXmC8X/+l8TIE7uBQ5FlA==",
+      "requires": {
+        "prop-types": "15.6.1",
+        "rip-out": "1.0.0"
+      }
+    },
```
tanx: @valentinewallace I removed the changes and rebased the branch.

tanx:
Also, I was getting a warning in the cobsole in storybook.

package.json
valentinewallace: Gotcha, i'll switch versions

valentinewallace:
> Also, I was getting a warning in the cobsole in storybook.

I see that you've fixed it? 
Let me know if there are other desired changes!

valentinewallace:
Plan is to move the `LoadNetworkSpinner` into the file for [this](https://projects.invisionapp.com/d/main/default/#/console/13241316/281903500/preview) screen once this PR is finalized, since it's custom-made for that screen

src/component/spinner.js
```diff
@@ -0,0 +1,206 @@
...
+//
+
+const smallStyles = StyleSheet.create({
+  spinner: {
+    transform: [{ scale: 1.42 }],
+  },
+});
+export const SmallSpinner = ({ ...props }) => (
+  <ActivityIndicator
+    sizeM="small"
```
tanx: @valentinewallace this prop does not exist. Only `size` https://facebook.github.io/react-native/docs/activityindicator#size

valentinewallace:
@tanx could we close this in favor of #406? it includes the same code but rebasing on top would be annoying because i moved the `LoadNetworkSpinner` into the file specific to its screen

PR #380, opened 2018-06-11
Notification list

### LINKED ISSUE #266: Implement Notifications Screen
https://projects.invisionapp.com/d/main/default/#/console/13241316/281949901/preview

tanx:
Closes #266

### DIFF
--- a/src/view/notification.js
+++ b/src/view/notification.js
@@ -1,0 +1,109 @@
+import React from 'react';
+import { View, StyleSheet } from 'react-native';
+import { observer } from 'mobx-react';
+import PropTypes from 'prop-types';
+import Background from '../component/background';
+import { Header, Title } from '../component/header';
+import { Button, BackButton, SmallPillButton } from '../component/button';
+import { ListContent, List, ListItem, ListHeader } from '../component/list';
+import { Alert } from '../component/notification';
+import Text from '../component/text';
+import { color, font } from '../component/style';
+
+//
+// Notification View
+//
+
+const NotificationView = ({ store, nav }) => {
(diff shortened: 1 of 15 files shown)


test/integration/action/action-integration.spec.js
```diff
@@ -131,7 +131,7 @@ describe('Action Integration Tests', function() {
     });
 
     nav1 = sinon.createStubInstance(NavAction);
-    notify1 = sinon.createStubInstance(NotificationAction);
+    notify1 = sinon.createStubInstance(NotificationAction, nav1);
```
valentinewallace: Wondering if instead of labeling the two sets of everything as 1 and 2, a more descriptive variable name (or a comment)? I'm a little unclear on what each set is being used for

src/view/notification.js
```diff
@@ -0,0 +1,109 @@
...
+    fontSize: font.sizeS,
+  },
+  alert: {
+    marginRight: 6,
+  },
+  group: {
+    flexDirection: 'row',
+    alignItems: 'center',
+  },
+  l: { flex: 8 },
```
valentinewallace: i'm a little confused by these variable names

test/integration/action/action-integration.spec.js
tanx: We just have two lnd nodes and respective action instances for each node to keep state. Kinda like two lightning apps connecting two each other just without the react ui.

src/view/notification.js
tanx: Just t-shirt sizes to set relative column width :)

PR #381, opened 2018-06-12
Closes #254: implement New Address screen

### LINKED ISSUE #254: Implement New Address Screen
https://projects.invisionapp.com/d/main/default/#/console/13241316/282749289/preview

valentinewallace:
Very similar to DepositView

### DIFF
--- a/src/view/new-address.js
+++ b/src/view/new-address.js
@@ -1,0 +1,86 @@
+import React from 'react';
+import { View, StyleSheet } from 'react-native';
+import { observer } from 'mobx-react';
+import PropTypes from 'prop-types';
+import { CopyText } from '../component/text';
+import { SplitBackground } from '../component/background';
+import MainContent from '../component/main-content';
+import { H1Text } from '../component/text';
+import { CopyButton, GlasButton } from '../component/button';
+import { CopiedNotification } from '../component/notification';
+import QRCode from '../component/qrcode';
+import { color } from '../component/style';
+
+const styles = StyleSheet.create({
+  qrWrapper: {
+    position: 'absolute',
+    top: 0,
+    left: 0,
+    right: 0,
+    bottom: 0,
+    justifyContent: 'center',
+    alignItems: 'center',
+  },
+  title: {
+    alignSelf: 'center',
+    marginTop: 30,
+    textAlign: 'center',
+  },
+  content: {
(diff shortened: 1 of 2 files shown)


src/view/new-address.js
```diff
@@ -0,0 +1,100 @@
...
+  },
+});
+
+const NewAddressView = ({ store, invoice }) => (
+  <SplitBackground image="purple-gradient-bg" bottom={color.blackDark}>
+    <View style={styles.qrWrapper}>
+      <QRCode size={130}>{store.walletAddressUri}</QRCode>
+    </View>
+    <MainContent style={styles.content}>
+      <CopySection />
```
tanx: Wrapped the copy text in a CopiedSection like in the SeedView.

src/view/new-address.js
```diff
@@ -0,0 +1,100 @@
...
+    alignItems: 'center',
+  },
+  copyBtn: {
+    backgroundColor: color.blackDark,
+    marginBottom: 40,
+  },
+  copied: {
+    bottom: 180,
+  },
+});
```
tanx: Played around with the margins to make it more aligned with the design.

src/view/seed.js
```diff
@@ -38,10 +38,10 @@ const copyStyles = StyleSheet.create({
   },
   title: {
     textAlign: 'center',
-    marginBottom: 15,
   },
   copyTxt: {
     textAlign: 'center',
+    marginTop: 15,
```
tanx: This was just something I saw in the SeedView. We should always add `topMargin` to the following element. I think I added this the wrong way and you did it right here.

stories/screen.js
```diff
@@ -66,6 +67,7 @@ storiesOf('Screens', module)
   .add('Loader - First Time', () => <Loader />)
   .add('Seed', () => <Seed store={store} />)
   .add('Seed Success', () => <SeedSuccess />)
+  .add('New Address', () => <NewAddress store={store} invoice={invoice} />)
```
tanx: This is just nitpicking, but I reordered the screens so they are in line with the user journey. This way it's easier to follow in storybook.

src/view/seed.js
valentinewallace: ah yeah good to be consistent about that

PR #397, opened 2018-06-21
Implement Bitcoin Units view

### LINKED ISSUE #267: Implement Bitcoin Unit Screen
https://projects.invisionapp.com/d/main/default/#/console/13241316/281949904/preview

valentinewallace:
One note -- i noticed the line underneath the header for both `Settings` and `Bitcoin Units` views are missing. Should I make PR adding it to both? 

Closes #267


src/action/setting.js
```diff
@@ -0,0 +1,22 @@
+import { UNITS } from '../config';
+
+class SettingAction {
+  constructor(store) {
+    this._store = store;
+  }
+
+  //
+  // Bitcoin Unit actions
+  //
```
tanx: We should add an init method here to stay consistent with other actions see e.g. InvoiceAction.

src/action/setting.js
```diff
@@ -0,0 +1,22 @@
...
+
+  setBitcoinUnit(unit) {
```
tanx: We should use named parameters like in the other actions. So in this case `({ unit })`.

src/action/setting.js
```diff
@@ -0,0 +1,22 @@
...
+    if (this._store.settings.unit === unit) {
+      return;
```
tanx: This check shouldn't really be necessary. No harm done in setting the same value again or am I missing something?

src/action/setting.js
```diff
@@ -0,0 +1,22 @@
...
+    } else if (!(unit in UNITS)) {
+      return;
```
tanx: Probably safe to throw an error here, as it would indicate wrong usage of the api. Maybe best to do this check at the beginning of the function.

src/view/bitcoin-units.js
```diff
@@ -0,0 +1,119 @@
...
+    style={[
+      iStyles.selection,
+      { backgroundColor: selected ? color.lightPurple : color.blackDark },
+    ]}
+  />
+);
+
+ItemRadioBtn.propTypes = {
+  selected: PropTypes.bool.isRequired,
+};
```
tanx: Probably best to pull this component out and make it reusable e.g. for fiat currency selection.

src/view/bitcoin-units.js
```diff
@@ -0,0 +1,119 @@
...
+          type="bit"
+          selectedUnit={store.settings.unit}
+          onSelect={() => setting.setBitcoinUnit('bit')}
+        />
+        <UnitsListItem
+          name="Bitcoin"
+          type="btc"
+          selectedUnit={store.settings.unit}
+          onSelect={() => setting.setBitcoinUnit('btc')}
+        />
```
tanx: Probably best to wrap the three UnitsListItems in a View and center that like in the design.

image

src/action/setting.js
valentinewallace: i missed that, thanks!

tanx:
@valentinewallace heads up I rebased this on the master.

valentinewallace:
@tanx I fixed the test, so should be good to go on my end!

tanx:
@valentinewallace I refactored the code quite a bit and integrated the view into the app. Let me know what you think. Ready to merge from my side.

valentinewallace:
@tanx Looks great, I like the refactor making a `SettingUnit` reuseable component. merging

PR #408, opened 2018-06-27
Cleanup store

### LINKED ISSUE #407: Refactor mobx store init so constructor has no side effects
* Invoke computed modules in init
* Pull AsyncStorage read into `restore` function
* Use `new Store()` in unit tests

tanx:
Closes #407


test/unit/action/info.spec.js
```diff
@@ -15,8 +15,7 @@ describe('Action Info Unit Tests', () => {
   beforeEach(() => {
     sandbox = sinon.createSandbox({});
     sandbox.stub(logger);
-    useStrict(false);
-    store = observable({ lndReady: false });
+    store = new Store();
```
valentinewallace: This refactor is very easy on the eyes :100: makes test writing way easier as well

src/action/index.js
```diff
@@ -20,7 +20,8 @@ const ipcRenderer = window.ipcRenderer; // exposed to sandbox via preload.js
 // Inject dependencies
 //
 
-store.init(AsyncStorage);
+store.init();
```
valentinewallace: This breaking-up of functions makes sense to me

src/store.js
```diff
@@ -33,32 +33,32 @@ export class Store {
         passwordVerify: '',
         seedVerify: ['', '', ''],
       },
-      transactions: null,
+      transactions: [],
```
valentinewallace: these look great! interesting we weren't initializing with empty lists from the beginning :O

tanx:
Awesome. Thanks for the feedback.

PR #420, opened 2018-06-29
Cli logs

### LINKED ISSUE #268: Implement just the Logs part of the CLI Screen
https://projects.invisionapp.com/d/main/default/#/console/13241316/281949903/preview

valentinewallace:
Close #268

### DIFF
--- a/src/action/log.js
+++ b/src/action/log.js
@@ -15,6 +15,9 @@ class LogAction {
     _ipcRenderer = ipcRenderer;
     _ipcRenderer.on('logs', (event, arg) => {
       store.logs.push(arg);
+      if (store.logs.length > 100) {
+        store.logs.splice(0, store.logs.length - 100);
+      }
     });
     _ipcRenderer.send('logs-ready', true);
   }
--- a/src/component/list.js
+++ b/src/component/list.js
@@ -53,9 +53,15 @@ export class List extends Component {
   render() {
     return (
       <ListView
+        ref={component => (this.list = component)}
         dataSource={this.dataSource}
         renderHeader={this.props.renderHeader}
-        renderRow={this.props.renderItem}
+        renderRow={data => {
+          this.props.scrollToEnd
+            ? setTimeout(() => this.list._scrollViewRef.scrollToEnd(), 50)
+            : null;
+          return this.props.renderItem(data);
+        }}
         enableEmptySections={true}
       />
     );
@@ -66,6 +72,7 @@ List.propTypes = {
   data: PropTypes.array.isRequired,
   renderHeader: PropTypes.func,
   renderItem: PropTypes.func.isRequired,
+  scrollToEnd: PropTypes.bool,
 };
 
 //
--- a/src/view/cli.js
+++ b/src/view/cli.js
@@ -1,10 +1,11 @@
 import React from 'react';
 import { StyleSheet } from 'react-native';
+import { observer } from 'mobx-react';
 import PropTypes from 'prop-types';
 import Background from '../component/background';
-import MainContent from '../component/main-content';
 import { Header, Title } from '../component/header';
 import Text from '../component/text';
+import { List, ListContent } from '../component/list';
 import { Button, BackButton } from '../component/button';
 import { color, font } from '../component/style';
 
@@ -29,9 +30,13 @@ const CLIView = ({ store, nav }) => (
       <Button disabled onPress={() => {}} />
     </Header>
     <Background color={color.cliBackground}>
-      <MainContent style={styles.content}>
-        {store.logs.map((log, i) => <LogItem text={log} key={i} />)}
-      </MainContent>
+      <ListContent>
+        <List
+          data={store.logs.slice()}
+          renderItem={text => <LogItem text={text} />}
+          scrollToEnd={true}
+        />
+      </ListContent>
     </Background>
   </Background>
 );
@@ -54,4 +59,4 @@ LogItem.propTypes = {
   text: PropTypes.string,
 };
 
-export default CLIView;
+export default observer(CLIView);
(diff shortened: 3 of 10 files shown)


src/view/cli.js
```diff
@@ -0,0 +1,57 @@
...
+CLIView.propTypes = {
+  store: PropTypes.object.isRequired,
+  nav: PropTypes.object.isRequired,
+};
+
+const iStyles = StyleSheet.create({
+  text: {
+    textAlign: 'left',
+    fontSize: font.sizeS,
+  },
```
tanx: I updated the fontSize here to be in line with the `SettingItem`.

src/view/cli.js
```diff
@@ -0,0 +1,57 @@
...
+  <Background color={color.blackDark} style={styles.wrapper}>
+    <Header separator>
+      <BackButton onPress={() => nav.goSettings()} />
+      <Title title="CLI" />
+      <Button disabled onPress={() => {}} />
+    </Header>
+    <Background color={color.cliBackground}>
+      <MainContent style={styles.content}>
+        {store.logs.map((log, i) => <LogItem text={log} key={i} />)}
+      </MainContent>
```
tanx: The expected behavior would be for the list to always scroll to the end like a terminal. React Native `ScrollView` and `ListView` both expose a `scrollToEnd` method: https://facebook.github.io/react-native/docs/listview#scrolltoend.

Here's how you would use that: https://stackoverflow.com/questions/42313308/react-native-listview-scrolltoend-it-doesnt-work

The `MainContent` wrapper uses a `ScrollView` under the hood which just renders all child components at once. Given that logs will probably become very long, using our `List` component will probably be faster. Might be worth testing with 100 LogItems and see how ScrollView performs.

In any case we should probably splice the `store.logs` array when items are pushed in the LogAction to limit to 100 max.

src/action/log.js
```diff
@@ -15,6 +15,9 @@ class LogAction {
     _ipcRenderer = ipcRenderer;
     _ipcRenderer.on('logs', (event, arg) => {
       store.logs.push(arg);
+      if (store.logs.length > 100) {
+        store.logs.splice(0, store.logs.length - 100);
+      }
```
tanx: This will return the first 100 not the last:
```
> ['foo', 'bar', 'baz', '42'].splice(0, 4 - 2)
returns [ 'foo', 'bar' ]
```

If we only had unit tests 😉

src/component/list.js
```diff
@@ -53,9 +53,15 @@ export class List extends Component {
...
+        ref={component => (this.list = component)}
         dataSource={this.dataSource}
         renderHeader={this.props.renderHeader}
-        renderRow={this.props.renderItem}
+        renderRow={data => {
+          this.props.scrollToEnd
+            ? setTimeout(() => this.list._scrollViewRef.scrollToEnd(), 50)
+            : null;
+          return this.props.renderItem(data);
+        }}
```
tanx: Can't we set the prop on the `ListView` itself? Seems a bit out of place here.

src/component/list.js
```diff
@@ -53,9 +53,15 @@ export class List extends Component {
   render() {
     return (
       <ListView
+        ref={component => (this.list = component)}
         dataSource={this.dataSource}
         renderHeader={this.props.renderHeader}
-        renderRow={this.props.renderItem}
+        renderRow={data => {
+          this.props.scrollToEnd
+            ? setTimeout(() => this.list._scrollViewRef.scrollToEnd(), 50)
```
tanx: Calling a private attribute will likely break once we upgrade react-native-web. Isn't there a public api for the ListView? Also, why call `_scrollViewRef` when it's a ListView not a ScrollView?

src/view/cli.js
```diff
@@ -0,0 +1,62 @@
...
+      <Button disabled onPress={() => {}} />
+    </Header>
+    <Background color={color.cliBackground}>
+      <ListContent>
+        <List
+          data={store.logs.slice()}
+          renderItem={text => <LogItem text={text} />}
+          scrollToEnd={true}
+        />
+      </ListContent>
```
tanx: This renders the scrollbar inside of the view. If we remove the `ListContent` and just add padding to the `LogItem` components it should look better.

src/view/cli.js
```diff
@@ -0,0 +1,62 @@
...
+// CLI View
+//
+
+const styles = StyleSheet.create({
+  content: {
+    alignItems: 'flex-start',
+    paddingBottom: 30,
+    paddingLeft: 50,
+    paddingRight: 50,
+  },
```
tanx: This style isn't used anywhere.

src/action/log.js
valentinewallace: Lol, point taken xD However the code is correct -- the return value of `splice` is the removed list items

src/component/list.js
valentinewallace: I agree, I'm looking for an elegant solution to this. There isn't a public API because the `react-native-web` `ListView` is feature-incomplete.

src/action/log.js
tanx: > the return value of splice is the items that were removed from the list

Ah ok. My bad. Well unit tests wouldn't hurt either way.

src/view/cli.js
tanx: I changed this back to use `ListContent` since it's more consistent with the way we did thing before.

valentinewallace:
@tanx cool to get this merged!

I was still working on making the scroll nicer. LMK if you want me to continue that or open an issue for it
[nicer = stays in position if the user scrolls up, like terminal]

tanx:
> I was still working on making the scroll behave more like terminal. LMK if you want me to continue that or open an issue for it.

Should be good enough for now. We can always revisit this later and optimize if users complain.  Unless you already spent a bunch of time on it?

valentinewallace:
> Should be good enough for now. We can always revisit this later and optimize if users complain. Unless you already spent a bunch of time on it?

No worries, just putting it out there! :)

PR #421, opened 2018-06-29
Lightning uri

### LINKED ISSUE #411: Implement url handler for lightning: links containing invoices
The url handler is already set in:
https://github.com/lightninglabs/lightning-app/blob/02aa961dad22b1eebafd27d24d71403eea7362a6/public/electron.js#L196-L201

We just need to verify the uri in a regex and handle the `open-url` ipc event in the rendering engine.

tanx:
Closes #411

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -181,6 +181,16 @@ export const reverse = src => {
 };
 
 /**
+ * Basic uri validation before rendering. More thorough matching
+ * is done by lnd. This is just to mitigates XSS.
+ * @param  {string}  str The uri to validate
+ * @return {boolean}     If the uri is valid
+ */
+export const isValidUri = str => {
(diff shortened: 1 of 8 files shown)


src/helper.js
```diff
@@ -180,6 +180,26 @@ export const reverse = src => {
...
+  return /^lightning:ln[a-zA-Z0-9]*$/.test(str);
+};
+
+/**
+ * Basic bitcoin address validation. More thorough matching is
+ * done by lnd. This is just to mitigate XSS.
+ * @param  {string}  str The address to validate
+ * @return {boolean}     If the uri is valid
+ */
+export const isAddress = str => {
```
valentinewallace: Ummm I believe we forgot litecoin support ;)

PR #422, opened 2018-06-29
Save settings

### LINKED ISSUE #231: Persist settings across app sessions

tanx:
Closes #231


test/unit/helper.spec.js
```diff
@@ -336,6 +340,12 @@ describe('Helpers Unit Tests', () => {
       const rate = helpers.calculateExchangeRate(100000, settings);
       expect(rate, 'to match', /6{1}[,.]9{1}0{1}/);
     });
+
+    it('should be infinite for unknown rate', () => {
+      settings.fiat = 'eur';
+      const rate = helpers.calculateExchangeRate(100000, settings);
+      expect(rate, 'to match', /∞/);
```
valentinewallace: Would this ever be the case, that the exchange rate is unknown?

test/unit/store.spec.js
```diff
@@ -0,0 +1,17 @@
+import { Store } from '../../src/store';
+
+describe('Store Unit Tests', () => {
+  let store;
+
+  beforeEach(() => {
+    store = new Store();
+  });
+
+  describe('init()', () => {
```
valentinewallace: nice :+1:

test/unit/action/app-storage.spec.js
```diff
@@ -0,0 +1,94 @@
+import { Store } from '../../../src/store';
+import AppStorage from '../../../src/action/app-storage';
+import * as logger from '../../../src/action/log';
+
+describe('Action App Storage Unit Tests', () => {
```
valentinewallace: :100:  tests

test/unit/helper.spec.js
tanx: No, but the user's machine might not have an internet connect to fetch a current exchange rate. So we need to handle this case and display something.

tanx:
> One thing I noticed was that I get a ton of repetitive "Syncing to chain..."notifications, I'm wondering if we should open a ticket for having a number next to repeated consecutive notifications rather than listing each notification individually

Good point. I was also hoping the CLI log screen would display the block height. But seems like we only display logs from lnd, not the app's info/error logs.

PR #437, opened 2018-07-05
Wait screen

### LINKED ISSUE #428: Display spinner screen after password input and before `lndReady` is true

tanx:
Just a super simple spinner while lnd is initializing for the alpha release. I'll ask Lee if he can design something nicer.

Closes #428

### DIFF
--- a/src/action/nav.js
+++ b/src/action/nav.js
@@ -31,6 +31,10 @@ class NavAction {
     this._store.route = 'NewAddress';
   }
 
+  goWait() {
+    this._store.route = 'Wait';
+  }
+
   goHome() {
     this._store.route = 'Home';
   }
--- a/src/view/wait.js
+++ b/src/view/wait.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import { StyleSheet, ActivityIndicator } from 'react-native';
+import Background from '../component/background';
+import MainContent from '../component/main-content';
+import { color } from '../component/style';
+
+const styles = StyleSheet.create({
+  content: {
+    justifyContent: 'center',
+  },
+  spinner: {
+    transform: [{ scale: 1.5 }],
+  },
+});
+
+const WaitView = () => (
+  <Background color={color.blackDark}>
+    <MainContent style={styles.content}>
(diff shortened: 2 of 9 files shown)


src/action/nav.js
```diff
@@ -31,6 +31,10 @@ class NavAction {
     this._store.route = 'NewAddress';
   }
 
+  goWait() {
```
valentinewallace: I guess it doesn't matter if it's really temporary, but I'd go with a more descriptive name like `waitForNetworkLoad`. Unless we plan on having 1 `Wait` screen for multiple types of waiting?

src/view/wait.js
```diff
@@ -0,0 +1,28 @@
...
+const styles = StyleSheet.create({
+  content: {
+    justifyContent: 'center',
+  },
+  spinner: {
+    transform: [{ scale: 1.5 }],
+  },
+});
+
+const WaitView = () => (
```
valentinewallace: Lol, looks good :ok_hand:

src/view/wait.js
```diff
@@ -0,0 +1,28 @@
...
+  <Background color={color.blackDark}>
+    <MainContent style={styles.content}>
+      <ActivityIndicator
+        size="large"
+        color={color.lightPurple}
+        style={styles.spinner}
+      />
```
valentinewallace: I might add a `copy` section with a small description of what the user is waiting on.

src/action/nav.js
tanx: Yes. I also added the wait screen for the lightning payment. That takes a few seconds on testnet

src/view/wait.js
tanx: Right. Engineer designed UI ftw :)

src/view/wait.js
tanx: Would make sense if we used it only on one screen. We're going to replace this anyway with a fancier design anyway. So I think it's ok for now.

tanx:
Merging for now. We can always iterate on the design once we have an update from Lee.

PR #446, opened 2018-07-10
Implement chain backend for Loader - Syncing chain

### LINKED ISSUE #255: Implement Loader - Syncing Chain Screen
https://projects.invisionapp.com/d/main/default/#/console/13241316/281903500/preview

valentinewallace:
Closes #255

### DIFF
--- a/src/action/index.js
+++ b/src/action/index.js
@@ -29,7 +29,7 @@ export const nav = new NavAction(store);
 export const grpc = new GrpcAction(store, ipcRenderer);
 export const notify = new NotificationAction(store, nav);
 export const wallet = new WalletAction(store, grpc, db, nav, notify);
-export const info = new InfoAction(store, grpc, notify);
+export const info = new InfoAction(store, grpc, nav, notify);
 export const channel = new ChannelAction(store, grpc, nav, notify);
 export const transaction = new TransactionAction(store, grpc, wallet, nav);
 export const invoice = new InvoiceAction(
@@ -72,10 +72,4 @@ observe(store, 'lndReady', () => {
   transaction.update();
   transaction.subscribeTransactions();
   transaction.subscribeInvoices();
-});
-
-observe(store, 'syncedToChain', () => {
-  if (store.route == 'LoaderSyncing') {
-    nav.goHome();
-  }
 });
--- a/src/action/index.js
+++ b/src/action/index.js
@@ -73,3 +73,9 @@ observe(store, 'lndReady', () => {
   transaction.subscribeTransactions();
   transaction.subscribeInvoices();
 });
+
+observe(store, 'syncedToChain', () => {
+  if (store.route == 'LoaderSyncing') {
+    nav.goHome();
+  }
+});
--- a/src/action/info.js
+++ b/src/action/info.js
@@ -17,6 +17,7 @@ class InfoAction {
       if (!response.synced_to_chain) {
         this._notification.display({ msg: 'Syncing to chain', wait: true });
         log.info(`Syncing to chain ... block height: ${response.block_height}`);
+        this._store.percentSynced = calcPercentSynced(response);
         clearTimeout(this.t3);
         this.t3 = setTimeout(() => this.getInfo(), RETRY_DELAY);
       }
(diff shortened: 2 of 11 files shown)


src/view/new-address.js
```diff
@@ -52,7 +52,13 @@ const NewAddressView = ({ store, nav, invoice }) => (
         >
           {store.walletAddress}
         </CopyButton>
-        <GlasButton onPress={() => nav.goHome()}>Done</GlasButton>
+        <GlasButton
+          onPress={() =>
+            store.syncedToChain ? nav.goHome() : nav.goLoaderSyncing()
+          }
```
tanx: We should keep all logic outside of react components since it isn't really testable. The info action might be a good place for this.

src/action/info.js
```diff
@@ -26,4 +27,11 @@ class InfoAction {
   }
 }
 
+const calcPercentSynced = response => {
+  const bestHeaderTimestamp = response.best_header_timestamp;
+  const currTimestamp = new Date().getTime() / 1000;
+  const percentSynced = bestHeaderTimestamp * 1.0 / currTimestamp;
+  return percentSynced;
+};
+
```
tanx: As mentioned, we should probably apply a function that approximates transactions over time, since the loader basically goes to 95% after a few seconds and stays there for minutes. @Roasbeef do you have an idea maybe?

src/action/index.js
```diff
@@ -73,3 +73,9 @@ observe(store, 'lndReady', () => {
   transaction.subscribeTransactions();
   transaction.subscribeInvoices();
 });
+
+observe(store, 'syncedToChain', () => {
+  if (store.route == 'LoaderSyncing') {
+    nav.goHome();
+  }
+});
```
tanx: If we put the navigation logic from the react component to an info action (see my comment below), we can do the observe there like in the wallet action: https://github.com/lightninglabs/lightning-app/blob/5545b3ee241a5df04a1894d328bfc583565c5b93/src/action/wallet.js#L130-L141

valentinewallace:
@tanx I switched up the spinner progress, IMO it looks better now. Lmk what you think!

test/unit/action/info.spec.js
```diff
@@ -16,11 +18,12 @@ describe('Action Info Unit Tests', () => {
     sandbox = sinon.createSandbox({});
     sandbox.stub(logger);
     store = new Store();
+    nav = sinon.createStubInstance(NavAction);
     require('../../../src/config').RETRY_DELAY = 1;
     grpc = sinon.createStubInstance(GrpcAction);
     grpc.sendCommand.resolves({});
     notification = sinon.createStubInstance(NotificationAction);
-    info = new InfoAction(store, grpc, notification);
+    info = new InfoAction(store, grpc, nav, notification);
```
tanx: You need to also add this argument in the `InfoAction` constructor in the integration tests.

test/unit/action/info.spec.js
```diff
@@ -50,6 +53,16 @@ describe('Action Info Unit Tests', () => {
...
+    it('should set percentSynced', async () => {
+      const testTimestamp = new Date().getTime();
+      grpc.sendCommand.withArgs('getInfo').resolves({
+        synced_to_chain: false,
+        best_header_timestamp: testTimestamp / 1000,
+      });
+      await info.getInfo();
+      expect(store.percentSynced, 'to be within', 0, 1);
+    });
+
```
tanx: This currently breaks the build: https://travis-ci.org/lightninglabs/lightning-app/jobs/403387961#L5266

src/action/info.js
```diff
@@ -14,16 +17,41 @@ class InfoAction {
       this._store.pubKey = response.identity_pubkey;
       this._store.syncedToChain = response.synced_to_chain;
       this._store.blockHeight = response.block_height;
+      if (this.firstTime) {
+        this.startingSyncTimestamp = response.best_header_timestamp || 0;
+        this.firstTime = false;
+      }
```
tanx: Couldn't we just query `if (!this.startingSyncTimestamp)`? Then we wouldn't need `this.firstTime`.

src/action/info.js
```diff
@@ -14,16 +17,41 @@ class InfoAction {
...
+        this._store.percentSynced = this.calcPercentSynced(response);
         clearTimeout(this.t3);
         this.t3 = setTimeout(() => this.getInfo(), RETRY_DELAY);
       }
     } catch (err) {
       log.error('Getting node info failed', err);
     }
   }
+
+  finishOnboarding() {
```
tanx: We should probably rename this to `initLoaderSyncing` to be consistent with the other actions.

PR #448, opened 2018-07-11
Display tx fee on lightning pay confirm view

### LINKED ISSUE #431: Display transaction fee in lightning payment confirmation screen
hit `QueryRoutes`, to the destination and amount, which will show the lower bound for a fee

valentinewallace:
Closes #431

### DIFF
--- a/src/action/payment.js
+++ b/src/action/payment.js
@@ -68,8 +68,13 @@ class PaymentAction {
       const request = await this._grpc.sendCommand('decodePayReq', {
         pay_req: invoice.replace(PREFIX_URI, ''),
       });
+      const fee = await this.estimateLightningFee({
+        destination: request.destination,
+        satAmt: request.num_satoshis,
+      });
       payment.amount = toAmount(parseSat(request.num_satoshis), settings.unit);
       payment.note = request.description;
+      payment.fee = toAmount(parseSat(fee), settings.unit);
       return true;
     } catch (err) {
       log.info(`Decoding payment request failed: ${err.message}`);
@@ -113,6 +118,20 @@ class PaymentAction {
       this._notification.display({ msg: 'Lightning payment failed!', err });
     }
     await this._transaction.update();
+  }
+
+  async estimateLightningFee({ destination, satAmt }) {
+    try {
+      const { routes } = await this._grpc.sendCommand('queryRoutes', {
+        pub_key: destination,
+        amt: satAmt,
+        num_routes: 1,
+      });
+      return routes[0].total_fees;
+    } catch (err) {
+      log.info(`Unable to retrieve fee estimate: ${err}`);
+      return 0;
+    }
   }
 }
 
--- a/src/action/payment.js
+++ b/src/action/payment.js
@@ -1,4 +1,4 @@
-import { PREFIX_URI } from '../config';
+import { PREFIX_URI, WAIT_DELAY } from '../config';
 import {
   toSatoshis,
   toAmount,
@@ -64,6 +64,7 @@ class PaymentAction {
 
   async decodeInvoice({ invoice }) {
     try {
+      const timer = setTimeout(() => this._nav.goWait(), WAIT_DELAY);
       const { payment, settings } = this._store;
       const request = await this._grpc.sendCommand('decodePayReq', {
         pay_req: invoice.replace(PREFIX_URI, ''),
@@ -75,9 +76,11 @@ class PaymentAction {
       payment.amount = toAmount(parseSat(request.num_satoshis), settings.unit);
       payment.note = request.description;
       payment.fee = toAmount(parseSat(fee), settings.unit);
+      clearTimeout(timer);
       return true;
     } catch (err) {
       log.info(`Decoding payment request failed: ${err.message}`);
+      this._nav.goPay();
       return false;
     }
   }
--- a/src/action/payment.js
+++ b/src/action/payment.js
@@ -129,7 +129,7 @@ class PaymentAction {
       });
       return routes[0].total_fees;
     } catch (err) {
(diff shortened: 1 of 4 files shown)


src/action/payment.js
```diff
@@ -114,6 +119,20 @@ class PaymentAction {
     }
     await this._transaction.update();
   }
+
+  async getFeeEstimate(destination, satAmt) {
```
tanx: We should probably rename this to `estimateLightningFee` as we will also have an o-chain `estimateFee`:
https://github.com/lightninglabs/lightning-app/pull/318/files

Also, all actions use paramter destructuring e.g. ({ param }) for all function. This makes it easier to read the code consuming the api and also has some other benefits. See:
http://2ality.com/2015/01/es6-destructuring.html#simulating-named-parameters-in-javascript

src/action/payment.js
```diff
@@ -114,6 +119,20 @@ class PaymentAction {
...
+    try {
+      const { routes } = await this._grpc.sendCommand('queryRoutes', {
+        pub_key: destination,
+        amt: satAmt,
+        num_routes: 1,
+      });
+      return routes[0].total_fees;
```
tanx: Noice 💯

test/unit/action/payment.spec.js
```diff
@@ -139,11 +139,16 @@ describe('Action Payments Unit Tests', () => {
...
+        destination: 'bar',
+      });
+      grpc.sendCommand.withArgs('queryRoutes').resolves({
+        routes: [{ total_fees: '100' }],
       });
       const isValid = await payment.decodeInvoice({ invoice: 'some-invoice' });
       expect(isValid, 'to be', true);
       expect(store.payment.amount, 'to match', /^0[,.]0{4}1{1}7{1}$/);
       expect(store.payment.note, 'to be', 'foo');
+      expect(store.payment.fee, 'to match', /^0[,.]0{5}1{1}$/);
```
tanx: In addition to the unit test, we should also add integration test.

src/action/payment.js
```diff
@@ -68,8 +68,13 @@ class PaymentAction {
...
         pay_req: invoice.replace(PREFIX_URI, ''),
       });
+      const fee = await this.estimateLightningFee({
+        destination: request.destination,
+        satAmt: request.num_satoshis,
+      });
       payment.amount = toAmount(parseSat(request.num_satoshis), settings.unit);
       payment.note = request.description;
+      payment.fee = toAmount(parseSat(fee), settings.unit);
       return true;
```
tanx: This would freeze the UI if `queryRoutes` takes a few seconds. Have you tested this with Yalls and Starblocks?

test/integration/action/action-integration.spec.js
```diff
@@ -344,10 +344,11 @@ describe('Action Integration Tests', function() {
       expect(isValid, 'to be', false);
     });
 
-    it('should decode invoice and return true', async () => {
+    it('should decode invoice, set fee and return true', async () => {
       const isValid = await payments1.decodeInvoice({
         invoice: store2.invoice.uri,
       });
+      expect(parseInt(store1.payment.fee), 'to be greater than or equal to', 0);
```
tanx: Heads up `payment.fee` is a string formatted float that is in the selected unit (BTC in this case). You'll wanna call `parseFloat(store1.payment.fee)` in this case.

src/action/payment.js
valentinewallace: Ya that's a good point, I did test with Yalls and Starblocks and it was fine but I'm adding a wait screen if `queryRoutes` takes more than 500ms.

src/action/payment.js
```diff
@@ -64,15 +64,23 @@ class PaymentAction {
 
   async decodeInvoice({ invoice }) {
     try {
+      const timer = setTimeout(() => this._nav.goWait(), WAIT_DELAY);
```
tanx: The user shouldn't need to wait to decode an invoice...

src/action/payment.js
```diff
@@ -64,15 +64,23 @@ class PaymentAction {
...
       const request = await this._grpc.sendCommand('decodePayReq', {
         pay_req: invoice.replace(PREFIX_URI, ''),
       });
+      const fee = await this.estimateLightningFee({
+        destination: request.destination,
+        satAmt: request.num_satoshis,
+      });
       payment.amount = toAmount(parseSat(request.num_satoshis), settings.unit);
       payment.note = request.description;
+      payment.fee = toAmount(parseSat(fee), settings.unit);
```
tanx: This function just decodes the invoice and returns a boolean.

I'd set `store.payment.fee` in `estimateLightningFee` itself and call it after `this._nav.goPayLightningConfirm()` here https://github.com/lightninglabs/lightning-app/pull/448/files#diff-c103796afd01e866f6af483b8930cf32R57.

src/action/payment.js
```diff
@@ -114,6 +122,20 @@ class PaymentAction {
...
+  async estimateLightningFee({ destination, satAmt }) {
+    try {
+      const { routes } = await this._grpc.sendCommand('queryRoutes', {
+        pub_key: destination,
+        amt: satAmt,
+        num_routes: 1,
+      });
+      return routes[0].total_fees;
+    } catch (err) {
+      log.info(`Unable to retrieve fee estimate: ${JSON.stringify(err)}`);
```
tanx: err.message should be enough here.

tanx:
@valentinewallace also a heads up... I rebased this onto the current master.

valentinewallace:
Lol omg I was almost ready to submit my (similar) updates on this one xD is there a way to avoid in the future? 

Looks good to me though :ok_hand: Thanks for adding the tests I missed :) I ran it in the packaged app and it worked great. Merging

PR #450, opened 2018-07-11
Even faster logs

### LINKED ISSUE #449: Throttle refresh rate of log output

tanx:
Closes #449

### DIFF
--- a/src/action/log.js
+++ b/src/action/log.js
@@ -21,10 +21,10 @@ export function error(...args) {
 
 function pushLogs(message) {
   if (!_store) return;
-  const { logs } = _store;
-  logs.push(message);
-  if (logs.length > MAX_LOG_LENGTH) {
-    logs.splice(0, logs.length - MAX_LOG_LENGTH);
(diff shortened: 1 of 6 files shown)


tanx:
> cool, as a JS dev i've never really done too much performance optimization so interesting to see in practice!
> 
> I'm sort of assuming this is mainly for the benefit of mobile? I didn't really notice any performance issues/latency before this change

The benefits of string concat vs. array join is minimal: https://jsperf.com/string-concat-vs-array-join-10000/15

The change is mainly to concat only once on each log event and not O(n) for every render of the view.

tanx:
> I'm sort of assuming this is mainly for the benefit of mobile? I didn't really notice any performance issues/latency before this change

After the initial syncing of the chain, lnd fires endless info logs while autopilot goes to town. The whole app basically freezes if the CLI screen is open.

tanx:
Alas, the biggest performance win was rendering the text in just a single react node, instead of a list of nodes (which each require the cost of a div in blink's/chromium rendering engine)

PR #455, opened 2018-07-17
Close channel/tx detail on escape press

### LINKED ISSUE #333: Close channel/transaction detail view when escape button is pressed

valentinewallace:
Closes #333

### DIFF
--- a/src/component/escapable.js
+++ b/src/component/escapable.js
@@ -2,15 +2,20 @@ import { Component } from 'react';
 import PropTypes from 'prop-types';
 
 class Escapable extends Component {
+  constructor(props) {
+    super(props);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+  }
+
   componentDidMount() {
-    if (document) {
-      document.addEventListener('keydown', e => this.handleKeyDown(e));
+    if (typeof document !== 'undefined') {
+      document.addEventListener('keydown', this.handleKeyDown);
     }
   }
 
   componentWillUnmount() {
-    if (document) {
-      document.removeEventListener('keydown', e => this.handleKeyDown(e));
+    if (typeof document !== 'undefined') {
+      document.removeEventListener('keydown', this.handleKeyDown);
     }
   }
 
--- a/src/component/escapable.js
+++ b/src/component/escapable.js
@@ -1,0 +1,28 @@
+import { Component } from 'react';
+import PropTypes from 'prop-types';
+
+class Escapable extends Component {
+  componentDidMount() {
+    if (document) {
+      document.addEventListener('keydown', e => this.handleKeyDown(e));
+    }
+  }
+
+  componentWillUnmount() {
+    if (document) {
+      document.removeEventListener('keydown', e => this.handleKeyDown(e));
+    }
+  }
+
+  handleKeyDown(event) {
+    if (event.keyCode === 27) {
+      this.props.onClose();
+    }
+  }
+}
+
+Escapable.propTypes = {
+  onClose: PropTypes.func.isRequired,
+};
+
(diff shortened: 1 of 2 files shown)


src/component/modal.js
```diff
@@ -34,15 +34,39 @@ const styles = StyleSheet.create({
...
-    {children}
-  </View>
-);
+class Modal extends Component {
+  constructor(props) {
+    props.title = props.title || '';
+    super(props);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+  }
+  componentDidMount() {
```
tanx: Missing empty line between the two functions.

src/component/modal.js
```diff
@@ -34,15 +34,39 @@ const styles = StyleSheet.create({
...
-      <Icon image="cancel-grey" style={styles.cancelIcon} />
-    </Button>
-    {children}
-  </View>
-);
+class Modal extends Component {
+  constructor(props) {
+    props.title = props.title || '';
+    super(props);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
```
tanx: Not a fan of the bind style. We should use arrow functions in the event handlers themselves to stay consistent with our current coding style.

src/component/modal.js
```diff
@@ -34,15 +34,39 @@ const styles = StyleSheet.create({
...
+  }
+  componentDidMount() {
+    document.addEventListener('keydown', this.handleKeyDown);
```
tanx: We should check if `document` exist as this will crash on mobile.

Also please use `e => this.handleKeyDown(e)` instead of bind.

src/component/modal.js
```diff
@@ -34,15 +34,39 @@ const styles = StyleSheet.create({
...
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.handleKeyDown);
+  }
```
tanx: Same as above...

src/component/modal.js
```diff
@@ -34,15 +34,39 @@ const styles = StyleSheet.create({
...
+      <View style={[styles.modal, style]}>
+        <H4Text style={styles.title}>{title.toUpperCase()}</H4Text>
+        <Button style={styles.cancelBtn} onPress={onClose}>
+          <Icon image="cancel-grey" style={styles.cancelIcon} />
+        </Button>
+        {children}
+      </View>
+    );
+  }
+}
```
tanx: I'm wondering if we should pull this out into an `Escapable` component that the modal inherits from.

src/component/escapable.js
```diff
@@ -0,0 +1,33 @@
...
+
+  componentDidMount() {
+    if (typeof document !== 'undefined') {
+      document.addEventListener('keydown', this.handleKeyDown);
+    }
+  }
+
+  componentWillUnmount() {
+    if (typeof document !== 'undefined') {
+      document.removeEventListener('keydown', this.handleKeyDown);
```
tanx: @valentinewallace it turn out we needed to use the bind style after all otherwise the function pointer would not have been consistent and the event listener would not have been removed.

src/component/escapable.js
```diff
@@ -0,0 +1,33 @@
...
+import PropTypes from 'prop-types';
+
+class Escapable extends Component {
+  constructor(props) {
+    super(props);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+  }
+
+  componentDidMount() {
+    if (typeof document !== 'undefined') {
```
tanx: When checking for a global like `document` or `window` always use `typeof` otherwise you'll get this error in an environment that does not have them: `ReferenceError: document is not defined`

PR #457, opened 2018-07-17
White list IPC events

### LINKED ISSUE #75: White list IPC calls in preload.js instead of directly exposing window.ipcRenderer
### Description

https://github.com/lightninglabs/lightning-app/blob/master/apps/desktop/main.dev.js#L128
NodeJS (nodeIntegration) is not disabled, this gives the JavaScript (loaded by the HTML files) in the renderer process full control over the system.

The Chromium sandbox is also disabled by default, I suggest enabling it but this may require architectural changes. I've created a comprehensive tutorial and boilerplate example on how to create [a secure electron application](https://github.com/kewde/electron-sandbox).

tanx:
Closes #75 

@kewde this removes direct access to `ipcRenderer` from the rendering process. Does this make sense or would you recommend further mitigation?

### DIFF
--- a/public/preload.js
+++ b/public/preload.js
@@ -1,2 +1,24 @@
-// expose only ipcRenderer to sandbox instead of all node apis
-window.ipcRenderer = require('electron').ipcRenderer;
+/**
+ * @fileOverview exposes only required apis to rendering process to minimize attack surface
+ */
+
+const _ipcRenderer = require('electron').ipcRenderer;
+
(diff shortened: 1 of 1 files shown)


kewde:
For bidirectional channels, the IPC is still the best route and I can't propose any further mitigation.

tanx:
Awesome! Thank for your insight. Very helpful.

PR #458, opened 2018-07-17
List only 1 'syncing to chain' ntfn

### LINKED ISSUE #456: List only a single "syncing to chain" notification but keep the notification bar visible until syncing completes

valentinewallace:
Closes #456

### DIFF
--- a/src/action/notification.js
+++ b/src/action/notification.js
@@ -9,15 +9,21 @@ class NotificationAction {
 
   display({ type, msg, wait, err, handler, handlerLbl }) {
     if (err) log.error(msg, err);
-    this._store.notifications.push({
-      type: type || (err ? 'error' : 'info'),
-      message: msg,
-      waiting: wait,
-      date: new Date(),
-      handler: handler || (err ? () => this._nav.goCLI() : null),
-      handlerLbl: handlerLbl || (err ? 'Show error logs' : null),
-      display: true,
-    });
+    const ntfnCount = this._store.notifications.length;
+    let prevNtfn = ntfnCount ? this._store.notifications[ntfnCount - 1] : null;
+    if (prevNtfn && prevNtfn.message === msg) {
+      prevNtfn.date = new Date();
+    } else {
+      this._store.notifications.push({
+        type: type || (err ? 'error' : 'info'),
+        message: msg,
+        waiting: wait,
+        date: new Date(),
+        handler: handler || (err ? () => this._nav.goCLI() : null),
+        handlerLbl: handlerLbl || (err ? 'Show error logs' : null),
+        display: true,
+      });
+    }
     clearTimeout(this.tdisplay);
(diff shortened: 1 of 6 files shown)


src/action/notification.js
```diff
@@ -9,15 +9,21 @@ class NotificationAction {
...
+      this._store.notifications.push({
+        type: type || (err ? 'error' : 'info'),
+        message: msg,
+        waiting: wait,
+        date: new Date(),
+        handler: handler || (err ? () => this._nav.goCLI() : null),
+        handlerLbl: handlerLbl || (err ? 'Show error logs' : null),
+        display: true,
+      });
+    }
```
tanx: This action should probably not need to be changed. Either we call `notification.display` only once in the `info` action (problem is we want the notification bar to stay visible, so this won't work) or we filter the notification list to render the syncing info only once. Both aren't  really solutions I'm happy with. Thoughts?

valentinewallace: I think it would make sense to modify `computedNotifications` to filter the notification list such that a new `Syncing to chain...` notification is shown every 5 minutes. I'd also be fine with only showing it once. what do you think?

tanx: Showing it once is fine. Filtering in computedNotifications makes sense. Something like...
```
const all = [];
notifications.forEach(n => {
  if (n.waiting && all.find(a => a.waiting)) {
    return;
  }
  all.push(n);
});
```

tanx: We only use the `waiting` flag for syncing so that should do the job.

src/computed/notification.js
```diff
@@ -13,7 +13,13 @@ const ComputedNotification = store => {
...
     computedNotifications: computed(() => {
       const { notifications } = store;
-      const all = notifications ? notifications.slice() : [];
+      const all = [];
+      notifications.forEach(n => {
+        if (n.waiting && all.find(a => a.waiting)) {
+          return;
+        }
+        all.push(n);
+      });
```
tanx: This displays only the first `waiting` notification.

src/computed/notification.js
```diff
@@ -22,6 +28,9 @@ const ComputedNotification = store => {
       });
       return all;
     }),
+    notificationCountLabel: computed(() =>
+      formatNumber(store.computedNotifications.length)
+    ),
```
tanx: Also moved this here to be computed based on the `computedNotifications` list size.

test/unit/computed/notification.spec.js
```diff
@@ -23,17 +24,38 @@ describe('Computed Notification Unit Tests', () => {
         date: new Date(1528703821406),
         display: true,
       });
+      store.notifications.push({
```
valentinewallace: Went in to add this test, I see you've already added it :100:

valentinewallace:
Everything great, tested packaged app & looks good :)

PR #459, opened 2018-07-18
Invoice note

### LINKED ISSUE #432: Display invoice description in lightning payment confirmation screen

tanx:
Closes #432

### DIFF
--- a/stories/screen.js
+++ b/stories/screen.js
@@ -200,6 +200,7 @@ store.payments = [...Array(10)].map(() => ({
   status: 'complete',
   date: new Date(),
 }));
+store.payment.note = '#craefulgang';
 store.selectedTransaction = (store.computedTransactions || []).find(
   tx => tx.type === 'bitcoin'
 );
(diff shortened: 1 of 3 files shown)


stories/screen.js
```diff
@@ -200,6 +200,7 @@ store.payments = [...Array(10)].map(() => ({
   status: 'complete',
   date: new Date(),
 }));
+store.payment.note = '#craefulgang';
```
valentinewallace: Added this so that the storybook shows a note

tanx: :)

tanx:
> Not sure what the policy is on me pushing commits to your branches, but I just added a message so that a note is shown in the storybook for pay-lightning-confirm.

That's fine. I think giving each other another chance to review the changes is a good idea. I have been a bit too optimistic on my end about that. Will try to give you a chance to review as well.

tanx:
> Tested the packaged app and it looks great! I'm wondering if there's a character limit that we should be abiding by or not

It's limited to one line for now and sets `...` if the string is longer.

PR #464, opened 2018-07-18
Fiat confirm

### LINKED ISSUE #460: Display lighting/bitcoin confirmation in fiat if that's selected
Already began work here: https://github.com/lightninglabs/lightning-app/tree/fiat-confirm

Just needs unit tests.

tanx:
Closes #460

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -127,6 +127,18 @@ export const toAmountLabel = (satoshis, settings) => {
 };
 
 /**
+ * Convert a string formatted BTC amount either to fiat or the selected BTC unit.
+ * The output should be used throughout the UI for value labels.
+ * @param  {string} amount The amount e.g. '0.0001'
+ * @param  {Object} settings Contains the current exchange rate
+ * @return {string}          The corresponding value label
+ */
+export const toLabel = (amount, settings) => {
+  const satoshis = toSatoshis(amount, settings.unit);
+  return toAmountLabel(satoshis, settings);
+};
+
+/**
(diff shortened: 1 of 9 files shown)


src/helper.js
```diff
@@ -126,6 +126,21 @@ export const toAmountLabel = (satoshis, settings) => {
...
 };
 
+/**
+ * Convert a string formatted BTC amount either to fiat or the selected BTC unit.
+ * The output should be used throughout the UI for value labels.
+ * @param  {string} amount The amount e.g. '0.0001'
+ * @param  {Object} settings Contains the current exchange rate
+ * @return {string}          The corresponding value label
+ */
+export const toLabel = (amount, settings) => {
```
valentinewallace: Would it make more sense for `toAmountLabel` and `toLabel` to be named `satsToAmountLabel` and `btcToAmountLabel` respectively instead?

test/unit/helper.spec.js
```diff
@@ -450,6 +450,66 @@ describe('Helpers Unit Tests', () => {
     });
   });
 
+  describe('toLabel()', () => {
```
valentinewallace: Thorough :1st_place_medal: :)

src/helper.js
tanx: > Would it make more sense for toAmountLabel and toLabel to be named satsToAmountLabel and btcToAmountLabel respectively instead?

Problem is the selected btc unit can be sat depending on the current settings.

Issue #292, opened 2018-04-04
Extract all strings into JSON file for en_US


tanx:

skv-headless:
Hello @tanx would love to help here.
Just want to suggest instead of keeping everything in 1 file we can keep messages inside components similar to what described here https://medium.freecodecamp.org/internationalization-in-react-7264738274a0. 
It looks a little bit ugly (https://github.com/iam-peekay/inbox-react-intl/) but seems to be a standart solution.

PR #467, opened 2018-07-21
[WIP] Locale setting 

torkelrogstad:
This PR adds a sub menu for selecting preferred user locale, as well as MobX state for said locale.

There's also a commit where the common container around the different settings views is made into a separate component, with the other settings sub menus updated to use the new container. 

This PR does not handle actually translating strings into the chosen locale, as that would need a decision on what approach to use. One option (which I've used on a React Native project without issue) is `i18n-js`. This PR can be left open until a decision is made and strings are extracted into JSON files (#292), and actual string translation is implemented.


tanx:
Thanks for the PR. We don't have any plans to add a locale setting. So closing this for now.

PR #471, opened 2018-07-26
Dont advertise chans

### LINKED ISSUE #469: Don't advertise channels when opening channels
Set bool on opening channel grpc

valentinewallace:
Closes #469

### DIFF
--- a/src/action/channel.js
+++ b/src/action/channel.js
@@ -61,8 +61,8 @@ class ChannelAction {
         remoteBalance: parseSat(channel.remote_balance),
         channelPoint: channel.channel_point,
         active: channel.active,
-        status: 'open',
         private: channel.private,
+        status: 'open',
       }));
     } catch (err) {
       log.error('Listing channels failed', err);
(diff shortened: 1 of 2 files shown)


test/integration/action/action-integration.spec.js
```diff
@@ -402,6 +402,13 @@ describe('Action Integration Tests', function() {
       expect(store1.computedChannels[0].status, 'to be', 'open');
     });
 
+    it('should have opened channels be private', async () => {
+      const { channels } = await grpc1.sendCommand('listChannels');
+      for (let channel of channels) {
+        expect(channel.private, 'to be', true);
+      }
+    });
+
```
tanx: Great that you added a check in the integration tests. The integration tests should stick to using higher level actions that are basically invoked by the UI, so we can kind of replicate the user flow without spinning up a DOM. I'd suggest just adding `expect(store1.computedChannels[0].private, 'to be', true);` to the previous test.

src/action/channel.js
```diff
@@ -162,6 +162,7 @@ class ChannelAction {
     const stream = this._grpc.sendStreamCommand('openChannel', {
       node_pubkey: new Buffer(pubkey, 'hex'),
       local_funding_amount: amount,
+      private: true,
```
tanx: Wohoo!

Issue #466, opened 2018-07-20
Implement Wait Screen redesign


tanx:
https://projects.invisionapp.com/share/UHN289CASW9#/screens/309738546_Loading_Spinner

PR #473, opened 2018-07-27
Implement Wait screen redesign

willpiers:
Addresses #466 

It doesn't seem like there are many hooks that we can listen to in order to make the spinner accurately depict progress - AFAIK the app is just issuing a `UnlockWallet` grpc command and waiting for a `lndReady` event, or something very similar.

This is why I created a continuous spinner

qcsk15df1k

Don't worry, the animation is smooth when not in gif format.

### DIFF
--- a/src/component/spinner.js
+++ b/src/component/spinner.js
@@ -1,5 +1,10 @@
-import React from 'react';
-import { ActivityIndicator, StyleSheet, View } from 'react-native';
+import React, { Component } from 'react';
+import {
+  ActivityIndicator,
+  StyleSheet,
+  View,
+  ViewPropTypes,
+} from 'react-native';
 import PropTypes from 'prop-types';
 import { color, font } from './style';
 import LightningBoltIcon from '../asset/icon/lightning-bolt';
@@ -43,7 +48,7 @@ export const LoadNetworkSpinner = ({ percentage, msg, style }) => (
     <ResizeableSpinner
       percentage={percentage}
       size={80}
-      progressWidth={3}
+      progressWidth={4}
       gradient="loadNetworkGrad"
     >
       <LightningBoltIcon height={28} width={14.2222} />
@@ -56,6 +61,43 @@ LoadNetworkSpinner.propTypes = {
   percentage: PropTypes.number.isRequired,
   msg: PropTypes.string.isRequired,
   style: View.propTypes.style,
+};
+
+//
+// ContinuousLoadNetworkSpinner
+//
+
+export class ContinuousLoadNetworkSpinner extends Component {
+  constructor(props) {
+    super(props);
+    this.increasePercentage = this.increasePercentage.bind(this);
+    const intervalId = setInterval(this.increasePercentage, 10);
+    this.state = {
+      percentage: 0,
+      intervalId,
+    };
+  }
+  render() {
+    const { msg, style } = this.props;
+    const { percentage } = this.state;
+    return (
+      <LoadNetworkSpinner msg={msg} percentage={percentage} style={style} />
+    );
+  }
+  increasePercentage() {
+    const { percentage } = this.state;
+    this.setState({
+      percentage: (percentage + 0.01) % 1,
+    });
+  }
+  componentWillUnmount() {
+    clearInterval(this.state.intervalId);
+  }
+}
+
+ContinuousLoadNetworkSpinner.propTypes = {
+  msg: PropTypes.string.isRequired,
+  style: ViewPropTypes.style,
 };
 
 //
@@ -132,10 +174,13 @@ const Gradients = () => (
 
 const SpinnerProgress = ({ width, percentage, color }) => (
   <Path
-    d={`M${width / 2} ${width / 2} L${width / 2} 0 ${generateArc(
-      percentage,
-      width / 2
-    )} Z`}
+    d={generateArc(
+      width / 2,
+      width / 2,
+      width / 2,
+      percentage * 360,
+      ((percentage + 0.4) % 1) * 360
+    )}
     fill={color}
   />
 );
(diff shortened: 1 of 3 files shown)


willpiers:
ah crap, probably should have checked outstanding PRs

willpiers:
Was having trouble sleeping, and realized we needed to adjusted the svg! This question on stack overflow was useful and I ended up using the code in the top answer https://stackoverflow.com/questions/5736398/how-to-calculate-the-svg-path-for-an-arc-of-a-circle

willpiers:
Looked into timers (and global scope) in react native a bit today. The following three chunks of code made me feel safer that a) `global` and `window` are the same in RN, and b) timers should totally work on all platforms and they rely on platform specific native timing modules:
[`global === window // true`](https://github.com/facebook/react-native/blob/master/Libraries/Core/InitializeCore.js#L34-L36)
[timers are polyfilled](https://github.com/facebook/react-native/blob/d01ab66b47a173a62eef6261e2415f0619fefcbb/Libraries/Core/InitializeCore.js#L102-L109)
[timers rely on native timing modules](https://github.com/facebook/react-native/blob/52f431b4bb29062abd8ce20e01a4e60b47151a80/Libraries/Core/Timers/JSTimers.js#L267-L287)

wereHamster:
You should start the timer on component mount, not in the constructor.

src/component/spinner.js
```diff
@@ -47,12 +47,13 @@ const loadNetworkStyles = StyleSheet.create({
...
 
-export const LoadNetworkSpinner = ({ percentage, msg, style }) => (
+export const LoadNetworkSpinner = ({ continuous, percentage, msg, style }) => (
   <View style={style}>
     <ResizeableSpinner
+      continuous={continuous}
       percentage={percentage}
       size={80}
-      progressWidth={3}
+      progressWidth={4}
```
tanx: 4 looks a bit too thick compared to the design. I'd stick with 3.

src/component/spinner.js
```diff
@@ -62,11 +63,55 @@ export const LoadNetworkSpinner = ({ percentage, msg, style }) => (
...
+        continuous={true}
+      />
+    );
+  }
+  increasePercentage() {
+    const { percentage } = this.state;
+    this.setState({
+      percentage: (percentage + 0.01) % 1,
+    });
+  }
```
tanx: Sorry if this didn't come across in the design. But the percentage should stay 25% statically for this spinner as it's just a generic wait screen that is used without any progress.

src/component/spinner.js
```diff
@@ -84,6 +129,7 @@ const resizeableStyles = StyleSheet.create({
 });
 
 export const ResizeableSpinner = ({
+  continuous,
```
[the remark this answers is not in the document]
willpiers: `continuous` needed to be passed into `SpinnerProgress`. Since `ResizableSpinner` sits between the top level spinner and the `SpinnerProgress` components, it has to go through `ResizableSpinner`

src/component/spinner.js
willpiers: okay cool. i noticed the startup spinner didn't look quite right but hadn't dug in to see what was going on.

willpiers: do you mean that this line should be `25` instead of `store.percentSynced`? https://github.com/lightninglabs/lightning-app/blob/master/src/view/loader-syncing.js#L24

src/component/spinner.js
```diff
@@ -139,17 +187,21 @@ const Gradients = () => (
...
-      percentage,
-      width / 2
-    )} Z`}
+    d={generateArc(
+      width / 2,
+      width / 2,
+      width / 2,
+      continuous ? percentage * 360 : 0,
+      continuous ? ((percentage + 0.4) % 1) * 360 : percentage * 360
+    )}
```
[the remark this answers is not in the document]
valentinewallace: I think you're redundantly taking the modulus in line `197`, could change to: 
` continuous ? (percentage + 0.4) * 360 : percentage * 360`

src/component/spinner.js
tanx: My mistake that as the `LoaderView`. My point was that you should be able just set `percentage={0.25}` on the `LoadNetworkSpinner` for the `ContinuousLoadNetworkSpinner` and get the desired result.

src/component/spinner.js
```diff
@@ -62,11 +63,55 @@ export const LoadNetworkSpinner = ({ percentage, msg, style }) => (
...
+
+export class ContinuousLoadNetworkSpinner extends Component {
+  constructor(props) {
+    super(props);
+    this.increasePercentage = this.increasePercentage.bind(this);
+    this.state = {
+      percentage: 0,
+    };
+  }
+  componentDidMount() {
```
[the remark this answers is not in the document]
willpiers: there is currently a prettier plugin being used by eslint https://github.com/lightninglabs/lightning-app/blob/master/.eslintrc
looks like it's just using the default prettier options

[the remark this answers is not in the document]
wereHamster: Relying on default options is dangerous, because people may have different default settings configured in their text editor / IDE. It's more sustainable to provide a .prettierrc in the repo itself.

tanx:
Hey. Would you mind rebaseing on the current master. I made some changes to the prop-types in a bunch of components. Thanks

src/component/spinner.js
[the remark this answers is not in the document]
tanx: > newlines between functions :)

Agreed. `.prettierrc` is out of scope for this PR. Patches welcome.

src/component/spinner.js
```diff
@@ -174,26 +226,25 @@ SpinnerFill.propTypes = {
...
+const polarToCartesian = (centerX, centerY, radius, angleInDegrees) => {
+  const angleInRadians = (angleInDegrees - 90) * Math.PI / 180.0;
+
+  return {
+    x: centerX + radius * Math.cos(angleInRadians),
+    y: centerY + radius * Math.sin(angleInRadians),
+  };
+};
+
+const generateArc = (x, y, radius, startAngle, endAngle) => {
```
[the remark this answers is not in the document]
tanx: We should also pull this logic out of the react components into `computed/loader.js` which we can unit test.

willpiers:
I will put some work in on this tomorrow. I've been quite busy, but would love to get this merged this week

src/component/spinner.js
```diff
@@ -165,26 +226,48 @@ SpinnerFill.propTypes = {
...
+/**
+ * Create an SVG path string for a circular arc
+ * @param  {number} x          The X-coord of the center of the circle
+ * @param  {number} y          The Y-coord of the center of the circle
+ * @param  {number} radius     The radius of the circle
+ * @param  {number} startAngle The angle at which to start the arc
+ * @param  {number} endAngle   The angle at which to end the arc
+ * @return {string}            The path string for the arc
+ */
+const generateArc = (x, y, radius, startAngle, endAngle) => {
```
willpiers: I would be happy to move these functions into a helper file so they are more easily unit tested. I don't really see the point of putting them in a `/computed/spinner` file since they don't currently rely on anything in the store (I know one of the spinner does use `store.percentSynced` but that doesn't touch this code path) and that would just seems like over-engineering to me.

tanx: > they don't currently rely on anything in the store

That's a valid point. My main concern is readaility and separation of concerns. Until now, we've been strict about keeping JS logic out and keeping the react components dumb. This way we don't need to spin up a dom or do testing on the component level, since all logic is already unit tested at the `computed` and `action` level.

erkarl: +1 for moving this into helper / util file and unit testing it.

tanx:
@willpiers any update on this PR 🙃

src/component/spinner.js
willpiers: I'm still not sure what exactly needs to be done to clear up this thread of comments. I'll move the two arc generation functions into a helper file, and write a couple tests. But if there is something else you need RE: this thread, either be overly specific so I know what to do, or maybe just wait til I'm done and add your own commit

tanx: Ok thanks @willpiers. Sounds good. Let me know when you're ready with your changes.

PR #474, opened 2018-07-27
Fix pending deposit

### LINKED ISSUE #468: Display correct pending deposit on home screen
We should probably add `pending_open_balance` to pending deposit...

@roasbeef:
noticed ealier that there may be a gap in how we display balances to the user atm
as we don't display the pending balance from the call: https://api.lightning.community/#channelbalance
so if we open a channel, and it hasn't yet been confirmed, then the two balances we display won't add up atm
by default the total_balance in wallet balance will sum the two 
but the balance that's from ChanneBalance actually won't as we don't display a sum of the two

tanx:
Closes #468

### DIFF
--- a/test/integration/action/action-integration.spec.js
+++ b/test/integration/action/action-integration.spec.js
@@ -307,7 +307,7 @@ describe('Action Integration Tests', function() {
       expect(store1.computedChannels[0].status, 'to be', 'pending-open');
     });
 
-    it('should have enough satoshis in channel balance', async () => {
+    it('should have enough satoshis in pending balance', async () => {
       await updateBalances();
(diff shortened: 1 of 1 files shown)


src/view/home.js
```diff
@@ -56,7 +56,7 @@ const HomeView = ({
       <QrCodeSeparator goDeposit={() => nav.goDeposit()} />
       <MainContent style={styles.content}>
         <BalanceDisplay
-          balanceLabel={balanceLabel}
+          depositLabel={depositLabel}
```
valentinewallace: suggestion: `depositLabel` --> `pendingDepositLabel`?

tanx: If there were multiple depositLabels. But short names are good names :)

PR #478, opened 2018-07-30
Fiat default

### LINKED ISSUE #477: Display fiat by default

tanx:
Closes #477

### DIFF
--- a/test/unit/store.spec.js
+++ b/test/unit/store.spec.js
@@ -9,9 +9,9 @@ describe('Store Unit Tests', () => {
 
   describe('init()', () => {
     it('should set computed vales', () => {
-      expect(store.unitLabel, 'to be', undefined);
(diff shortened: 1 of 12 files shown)


test/unit/store.spec.js
```diff
@@ -9,9 +9,9 @@ describe('Store Unit Tests', () => {
 
   describe('init()', () => {
     it('should set computed vales', () => {
-      expect(store.unitLabel, 'to be', undefined);
+      expect(store.unitFiatLabel, 'to be', undefined);
```
valentinewallace: should we be testing all of the computed values, or is this just to test that `init` changes this specific value? 

i guess if we tested all of them it would be redundant to the `ComputedWallet` tests.

tanx: Yeah this is just a smoke test of init.

PR #489, opened 2018-08-02
Add assets/bin to gitignore

erkarl:
Following the "build packaged app" from README includes `cp
$GOPATH/bin/lnd ./assets/bin/darwin` which will create an untracked
`assets/bin/darwin/lnd`.


tanx:
@erkarl thanks. LGTM

PR #490, opened 2018-08-03
Detect locale

### LINKED ISSUE #486: Detect currency based on locale
Electron has an api to detect the current locale: https://electronjs.org/docs/api/locales

N.B. this api needs to be called after the `ready` event in the main process to work correctly.

We can then send the locale string via IPC to the rendering process and detect the local currency using:
https://github.com/tadeegan/locale-currency

willpiers:
Closes #486

### DIFF
--- a/public/grpc-client.js
+++ b/public/grpc-client.js
@@ -44,7 +44,6 @@ module.exports.init = async function({
   ipcMain,
   lndPort,
   lndSettingsDir,
-  locale,
   macaroonsEnabled,
 }) {
   let credentials;
@@ -53,10 +52,6 @@ module.exports.init = async function({
   let lnrpc;
   let unlocker;
   let lnd;
-
-  ipcMain.on('get-locale', event => {
-    event.sender.send('locale', { response: { locale } });
-  });
 
   ipcMain.on('unlockInit', async event => {
     credentials = await getCredentials(lndSettingsDir);
--- a/public/grpc-client.js
+++ b/public/grpc-client.js
@@ -44,6 +44,7 @@ module.exports.init = async function({
   ipcMain,
   lndPort,
   lndSettingsDir,
+  locale,
   macaroonsEnabled,
 }) {
   let credentials;
@@ -52,6 +53,10 @@ module.exports.init = async function({
   let lnrpc;
   let unlocker;
   let lnd;
+
+  ipcMain.on('get-locale', event => {
+    event.sender.send('locale', { response: { locale } });
+  });
 
   ipcMain.on('unlockInit', async event => {
     credentials = await getCredentials(lndSettingsDir);
--- a/src/action/grpc.js
+++ b/src/action/grpc.js
@@ -13,6 +13,11 @@ class GrpcAction {
     this._ipcRenderer = ipcRenderer;
   }
 
+  async getLocale() {
+    let response = await this._sendIpc('get-locale', 'locale');
+    return response.locale;
+  }
+
   //
   // WalletUnlocker grpc client
   //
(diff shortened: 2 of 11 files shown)


public/grpc-client.js
```diff
@@ -53,6 +54,10 @@ module.exports.init = async function({
   let unlocker;
   let lnd;
 
+  ipcMain.on('get-locale', event => {
+    event.sender.send('locale', { response: { locale } });
+  });
```
willpiers: wasn't sure if these lines should be here or in `electron.js` - feedback wanted

src/action/setting.js
```diff
@@ -16,13 +18,14 @@ class SettingAction {
...
-    }
-    this._store.settings.fiat = fiat;
+    this._store.settings.fiat = FIATS[fiat] ? fiat : DEFAULT_FIAT;
     this._wallet.getExchangeRate();
     this._db.save();
   }
+  setFiatByLocale({ locale }) {
+    const fiat = LocaleCurrency.getCurrency(locale).toLowerCase();
+    this.setFiatCurrency({ fiat });
+  }
```
tanx: ... `if (!FIATS[fiat])` falling back to `usd` if the app does not support the currency yet. This will allow us to detect Euro and Pounds for now and add more currency support in the future (which requires a design change in the settings).

test/integration/action/action-integration.spec.js
```diff
@@ -203,6 +210,19 @@ describe('Action Integration Tests', function() {
...
+      expect(locale1, 'to be', 'gb');
+      settings1.setFiatByLocale({ locale: locale1 });
+      expect(store1.settings.fiat, 'to be', 'gbp');
+
+      const locale2 = await grpc2.getLocale(); // "Shona (Zimbabwe) - sn-ZW"
+      settings2.setFiatByLocale({ locale: locale2 });
+      expect(store2.settings.fiat, 'to be', 'usd');
+    });
+  });
+
```
tanx: No need to add an integration test since we don't spin up electron here anyway. Unit tests should suffice for this feature.

src/action/setting.js
willpiers: the ternary expression on line 21 does just that - if `fiat` is any of our supported fiats, it will be used. otherwise USD

src/action/grpc.js
```diff
@@ -84,6 +84,17 @@ class GrpcAction {
...
+  /**
+   * Wrapper function to execute a call to the main process
+   * in order to retrieve the user's locale.
+   * @param  {string} method The lnd GRPC api to call
+   * @param  {Object} body   The payload passed to the api
+   * @return {Promise<Object>}
+   */
+  sendLocaleRequest() {
+    return this._sendIpc('get-locale', 'locale');
+  }
```
willpiers: options here:
1) make `_sendIpc` public so it can be called from `actions/setting`
2) replicate logic of `_sendIpc` in `actions/setting`, and pass `ipcRenderer` into that file
3) leave this code here, which seems out of place

I would prefer 1 or 3, just for aesthetics and simplicity

src/action/setting.js
```diff
@@ -31,13 +34,16 @@ class SettingAction {
...
+    this._store.settings.fiat = FIATS[fiat] ? fiat : DEFAULT_FIAT;
     this._wallet.getExchangeRate();
     this._db.save();
   }
+
+  async getLocale() {
+    let response = await this._grpc.sendLocaleRequest();
+    const fiat = LocaleCurrency.getCurrency(response.locale).toLowerCase();
+    this.setFiatCurrency({ fiat });
+  }
```
willpiers: i'll add an annotation to this function

tanx:
Closing in favor of #501

PR #521, opened 2018-08-15
Inline svg

### LINKED ISSUE #289: Inline SVGs in a platform independent way
I've looked into how we can leverage the `svgr` compiler in order to inline SVGs into react components. This is required because the react-native `Image` component doesn't support svg files on iOS and Android. Rather we need to use `react-native-svg` to draw SVG using native primitives. In order to leverage the same `react-native-svg` components in electron together with react-native-web, I've created a PR to `create-react-app`:

https://github.com/facebook/create-react-app/pull/4224

Once this PR is merged we should be able to use a single SVG format on all platforms. Inlining SVGs also has the advantage that we can style them using CSS and don't need to edit colors in Sketch.

I've already gotten the `svgr` compiler to work on a branch: https://github.com/lightninglabs/lightning-app/commit/7ce81a26a181d096299b2f73ee41856715f4138a

tanx:
Closes #289


src/view/welcome.js
```diff
@@ -29,8 +24,12 @@ const styles = StyleSheet.create({
 
 const WelcomeView = () => (
   <Background image="textured-bg" style={styles.background}>
-    <Icon image="lightning-bolt" style={styles.bolt} />
-    <Icon image="lightning-word" style={styles.name} />
+    <View style={styles.boltWrapper}>
```
valentinewallace: can a `style` prop be passed into an `Svg` component to avoid using these wrappers?

tanx: Yeah, I tried that. Unfortunately the root `Svg` nodes are mapped to `svg` in the browser and those can't have a style attribute. They can on mobile however. So unfortunately the only cross-platform workaround is wrapping in a View 😕

Issue #424, opened 2018-07-02
lightning: uris aren't registered on Linux


tanx:

torkelrogstad:
I've done some research on this topic, and it looks like this isn't possible (at least not in a convenient way) while developing the app. By looking at some other apps (for example [`webtorrent-desktop`](https://github.com/webtorrent/webtorrent-desktop)), it looks like this is done when installing the app, at least on Ubuntu (what I'm using). After installing the app, clicking on magnet links work, but not while in dev mode. I'll try to use `electron-builder` to generate a `.deb` package with an accompanying `.desktop`-file, and see if that solves the issue. I guess that's somewhat related to #375, at least the package bit.

valentinewallace:
@torkelrogstad very cool, i'll take a look at your PR soon!

Issue #375, opened 2018-06-08
Create gpg signed deb package for Debian/Ubuntu


tanx:
See: https://github.com/signalapp/Signal-Desktop/blob/development/package.json#L195

And https://blog.packagecloud.io/eng/2014/10/28/howto-gpg-sign-verify-deb-packages-apt-repositories/

tanx:
This is how Signal is installed on Debian (https://signal.org/download/):
```
curl -s https://updates.signal.org/desktop/apt/keys.asc | sudo apt-key add -
echo "deb [arch=amd64] https://updates.signal.org/desktop/apt xenial main" | sudo tee -a /etc/apt/sources.list.d/signal-xenial.list
sudo apt update && sudo apt install signal-desktop
```

torkelrogstad:
```bash
$ curl -s https://packagecloud.io/install/repositories/torkelrogstad/lightning-app/script.deb.sh | sudo bash
$ sudo apt-get install lightning-app=0.2.0-prealpha.21
```
:slightly_smiling_face: 

As per #523: 
After spending a couple of hours doing research on different solutions (Launchpad.net, self hosted PPA, etc), a hosted solution like Packagecloud.io seems like the best way to do it. It looks like it's easy to set up Travis auto push ([docs](https://docs.travis-ci.com/user/deployment/packagecloud/)), it's not too expensive, and they've also got some sort of possibility for open source software using their services for free. As far as I can tell, they implement automatic GPG signing. 

By judging from the output of my `apt`, Slack does their Debian package handling this way:
```bash
$ sudo apt update
...
Hit:19 https://packagecloud.io/slacktechnologies/slack/debian jessie InRelease
```

tanx:
@torkelrogstad cool!

> As far as I can tell, they implement automatic GPG signing.

Is it possible to sign the deb bundle on travis or does the signing key need to resign with `Packagecloud.io`?

tanx:
> After spending a couple of hours doing research on different solutions (Launchpad.net, self hosted PPA, etc), a hosted solution like Packagecloud.io seems like the best way to do it.

Bitcoin Core also hosts on Launchpad. Isn't that also hosted by Canonical?

torkelrogstad:
> Is it possible to sign the deb bundle on travis or does the signing key need to reside with Packagecloud.io?

[...]

PR #523, opened 2018-08-15
Add .deb, improve Linux build

torkelrogstad:
Addressses #424 and #375.

Tweaks package.json to allow .deb building, adds custom mimetype for Linux URI support.

Author and homepage fields in `package.json` had to be tweaked for `electron-builder` to be happy about building a deb-package. I chose what seemed like sensible values, but feel free to change. I tested that pressing on a `lightning://foo` opened the app, but I'm not quite sure how this can be handled by tests. Is that an issue?

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,8 +2,12 @@
   "name": "lightning-app",
   "version": "0.2.0-prealpha.21",
   "description": "Lightning Wallet Application",
-  "author": "Lightning Labs, Inc",
-  "homepage": "./",
+  "author": {
+    "name": "Lightning Labs, Inc",
+    "email": "hello@lightning.engineering",
+    "url": "https://lightning.engineering"
+  },
+  "homepage": "https://lightning.engineering",
   "license": "GPL-3.0",
   "private": true,
   "main": "public/electron.js",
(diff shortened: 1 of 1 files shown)


tanx:
@torkelrogstad thanks for the PR.

Re: #424 It would be helpful to adress this issue in a separate PR so that we can test/integrate that first.

Re: #375 Unfortunately there's a bunch of other infrastructure that goes into getting signed .deb deployment working. Not the least of which is configuring GPG (see the issue description for details). I would recommend in looking into that as a second step.

torkelrogstad:
I think you can view this PR as a separate PR for #424. From what I've understood about handling URIs in Linux, it's done when installing the app. So in order to handle URIs, you have to make a deb package or similar.

tanx:
> From what I've understood about handling URIs in Linux, it's done when installing the app. So in order to handle URIs, you have to make a deb package or similar.

Interesting. Have we verified that auto update works with `.deb` installs? From what I gather you need to set up a PPA which would depend on #375.

torkelrogstad:
It sounds like #375 should be done first, then. I'll look into it.

Issue #480, opened 2018-07-31
Generate github release changelog from merged PRs


tanx:

willpiers:
I found a couple of nice projects for this kind of thing. But it depends what kind of process you'd want to have. Using [conventional commit message](https://conventionalcommits.org/) would help make a nice release notes. [Commitizen](https://github.com/commitizen/cz-cli) is really helpful with this.

The best option i've seen is [semantic-release](https://github.com/semantic-release/semantic-release). It's quite popular and well-maintained, and it has a simple enough [setup process](https://github.com/semantic-release/semantic-release/blob/caribou/docs/usage/ci-configuration.md#automatic-setup-with-semantic-release-cli)

tanx:
To clarify, we already have semantic versioning and CI builds with auto updates in place via travis/electron-builder/electron-updater. This issue is just to generate the change log from commit/PR messages and publish those in our github release notes: https://github.com/lightninglabs/lightning-app/releases

willpiers:
Yep, understood. My main suggestion is that standardizing commit messages would make it easier/possible to have release notes automatically generated.

erkarl:
As we don't have conventional commit messages in place I'm going to take a stab at implementing a primitive solution that just bundles all the commit messages together into an unsorted `CHANGELOG`.

erkarl:
Really primitive solution is ready for review: https://github.com/lightninglabs/lightning-app/pull/536

tanx:
Thanks @erkarl. Taking a look now.

erkarl:
@tanx there's also:
* https://github.com/github-tools/github-release-notes
* https://github.com/conventional-changelog/conventional-changelog

...and some others that I already ruled out due to old releases. Looking into those.

valentinewallace:
Was this not closed by https://github.com/lightninglabs/lightning-app/pull/536?

tanx:
No. The script didn’t work last time I tested it. @erkarl does it work on your fork when you run the script after deployment?

erkarl:
@tanx Sorry to hear it didn't work. I haven't setup Travis deployment on my fork. It successfully appended to a manually generated draft release, though.

PR #536, opened 2018-08-17
Generate release notes from commit messages

erkarl:
https://github.com/lightninglabs/lightning-app/issues/480

This PR will add a `generate-release-notes` script that will generate `CHANGELOG.md` based on new commits since the latest tag.

I didn't add automatic tag publishing because we might want to edit the release notes manually before publishing.


tanx:
@erkarl thanks for the PR. The main issue is updating the Github Release notes text via api. Perhaps there's a curl command that can be called with a `GH_TOKEN` env var?

erkarl:
Gotcha, I'll see what I can do about that.

erkarl:
@tanx I've updated the script to hit the GitHub releases endpoint to create a new draft release and include the commit messages as release notes.

screen shot 2018-08-21 at 16 29 35

Usage: `GH_TOKEN=123 ./release.sh v2.0`

`GH_TOKEN` can be generated from [github.com/settings/tokens](https://github.com/settings/tokens) and has to have `Full control of private repositories` permission.

release.sh
```diff
@@ -0,0 +1,41 @@
...
+TEMP_DATA_JSON="release.json"
+PACKAGE="package.json"
+GITHUB_API_ENDPOINT="https://api.github.com/repos/lightninglabs/lightning-app/releases"
+
+if [ $# -ne 1 ]; then
+    echo You must provide only a release version!
+    echo Current release: `sed -n 's/.*"version": "\(.*\)".*/\1/p' ${PACKAGE}`
+    exit 1
+fi
+RELEASE=$1
```
tanx: I don't fully understand this block. Should the version provided in the package.json be enough to identify the corresponding github release?

release.sh
```diff
@@ -0,0 +1,41 @@
...
+rm -f ${TEMP_DATA_JSON}
+
+DATA_JSON='{"tag_name": "'"$RELEASE"'", "name": "'"$RELEASE"'", "body": "'"$RELEASE_NOTES"'", "prerelease": true, "draft": true}'
+
+echo "${DATA_JSON}" >> ${TEMP_DATA_JSON}
+
+curl -H "Content-Type: application/json" \
+  -H "Authorization: token $GH_TOKEN" \
+  --data @$TEMP_DATA_JSON \
+  $GITHUB_API_ENDPOINT
```
tanx: Cool 👍

release.sh
tanx: > No, new version name must be provided as an argument. If you don't provide one the script will exit printing you the current version name for convenience.

Just to clarify, I run `npm version 0.2.2-alpha` locally and then push the git tag and updates package.json to the master. So the package.json already has to new version.

erkarl: Oh, thanks for clearing this up. I'll make the necessary changes. The script will then just take the version from `package.json` and generate a draft release notes to Github.

tanx: Cool. Sounds good. Thanks

release-notes.sh
```diff
@@ -0,0 +1,29 @@
...
+  exit 1
+fi
+
+TEMP_DATA_JSON="release.json"
+GITHUB_API_ENDPOINT="https://api.github.com/repos/lightninglabs/lightning-app/releases"
+RELEASE_NAME=`sed -n 's/.*"version": "\(.*\)".*/\1/p' package.json`
+CURRENT_TAG=v${RELEASE_NAME}
+PREVIOUS_TAG=$(git describe --tags --abbrev=0 $(git rev-list --tags --skip=1 --max-count=1))
+RELEASE_NOTES="$(git log ${PREVIOUS_TAG}..HEAD --merges --pretty='* %b [%h](http://github.com/lightninglabs/lightning-app/commit/%H)\n')"
+DATA_JSON='{"tag_name": "'"$CURRENT_TAG"'", "name": "'"$RELEASE_NAME"'", "body": "'"$RELEASE_NOTES"'", "prerelease": true, "draft": true}'
```
tanx: FYI we don't publish as `prerelease` anymore.

erkarl:
> Have you tried what happens when there is already a draft for the release there? Because the binaries should already have been pushed to the same draft before this new script get executed.

@tanx it looks like it currently creates a new draft with the same name. Ideally I'd like it to append to the existing draft.

erkarl:
@tanx I've updated the release-notes script to update an existing draft release generated by Travis. I got fed up with bash and refactored the script in JavaScript. Didn't use NodeJS `https` API because I already had the curl commands.

Usage: `node release-notes.js`

screen shot 2018-09-14 at 16 27 49

tanx:
> @tanx I've updated the release-notes script to update an existing draft release generated by Travis. I got fed up with bash and refactored the script in JavaScript. Didn't use NodeJS `https` API because I already had the curl commands.

Ok. Just curious about the choice of node since the code calls shell commands via node now. Did you run into some issue with bash?

erkarl:
@tanx In hindsight the shell script is probably a little easier to read. I've reverted back the changes and added the update existing release functionality to the shell script. For comparison the JS version is [available here](https://github.com/erkarl/lightning-app/blob/release-notes-js/release-notes.js).

PR #539, opened 2018-08-20
Update wallet on channel event

### LINKED ISSUE #510: Update wallet balances on channel open/close

tanx:
Closes #510

### DIFF
--- a/src/action/channel.js
+++ b/src/action/channel.js
@@ -7,9 +7,10 @@ import { toSatoshis, parseSat } from '../helper';
 import * as log from './log';
 
 class ChannelAction {
-  constructor(store, grpc, nav, notification) {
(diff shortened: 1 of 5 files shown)


src/action/channel.js
```diff
@@ -7,9 +7,10 @@ import { toSatoshis, parseSat } from '../helper';
 import * as log from './log';
 
 class ChannelAction {
-  constructor(store, grpc, nav, notification) {
+  constructor(store, grpc, transaction, nav, notification) {
```
erkarl: Out of curiosity - what's the logic behind adding a new param in the middle of existing ones rather than to the end? Also, in case of 2+ args I'm a proponent of passing in an object so that the ordering does not matter.

PR #541, opened 2018-08-20
Display funding Transaction ID in channel details

### LINKED ISSUE #529: show funding txid in channel details page
Atm, in the channel details page we show this: 
screen shot 2018-08-15 at 3 05 57 pm

The ID here is the short channel ID which is used internally within the protocol to allow nodes to discover new channels. However, it's mostly useless outside of it as most explorers don't understand this integer, and instead accept the txid's. As a result, we should display the _funding_ txid here instead. In listchannels, we show the `chan_point` which has the format: `txid:index`, so we can extract the txid with a bit of string manipulation. Note that the txid has already been reversed here, so it's ready for direct display.

tanx:
Closes #529

### DIFF
--- a/src/view/channel-detail.js
+++ b/src/view/channel-detail.js
@@ -34,16 +34,9 @@ const ChannelDetailView = ({ store, nav }) => (
         style={styles.modal}
         onClose={() => nav.goChannels()}
       >
-        {store.selectedChannel.id ? (
-          <DetailField name="Channel ID">
-            {store.selectedChannel.id}
-          </DetailField>
-        ) : null}
-        {store.selectedChannel.fundingTxId ? (
-          <DetailField name="Funding Transaction ID">
-            {store.selectedChannel.fundingTxId}
-          </DetailField>
-        ) : null}
+        <DetailField name="Funding Transaction ID">
+          {store.selectedChannel.fundingTxId}
+        </DetailField>
         <DetailField name="Remote Node Public Key">
           {store.selectedChannel.remotePubkey}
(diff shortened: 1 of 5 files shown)


src/view/channel-detail.js
```diff
@@ -34,7 +34,16 @@ const ChannelDetailView = ({ store, nav }) => (
         style={styles.modal}
         onClose={() => nav.goChannels()}
       >
-        <DetailField name="ID">{store.selectedChannel.id}</DetailField>
+        {store.selectedChannel.id ? (
```
erkarl: Can we love the ternary operator here? `{store.selectedChannel.id && ...`

tanx:
> Seems from the original issue #529 that the channel ID itself is pretty useless and we shouldn't display it at all.

Good catch. Yeah makes sense. I replaced channel id with the funding transaction id in the channel detail view and channel list.

> We could even display the funding transaction ID as "Channel ID" and skip showing the "real" channel ID?

I think `Funding Transaction ID` will help the user make the connection between the attribute on this screen and the one in the transaction list. Channel ID is gone now completely in the UI.

> Hm.. Additionally, I think it would be a good idea to show the closing tx ID when a channel is (pending) closed. Lmk if you agree and i'll open an issue! :)

Good point. Yeah we can tackle that in another issue.

erkarl:
LGTM!

valentinewallace:
I opened an issue for adding the closing tx ID. 

Changes look good! Merging.

PR #542, opened 2018-08-20
Display `Transaction ID` and `Invoice ID` in tx details

### LINKED ISSUE #526: rename ID to payment hash for lightning payments
For Lightning payments, we show the users this atm: 
screen shot 2018-08-15 at 3 10 35 pm

The ID nomenclature is a bit confusing here, as users _generally_ know this field as the "payment hash" 
and it may be used in the future to give a merchant an invoice ID for support, etc.

### LINKED ISSUE #527: rename ID to txid for on-chain payments
For on-chain payments atm we show the users this: 
screen shot 2018-08-15 at 3 10 46 pm

For _existing_ Bitcoin users, this can be a bit confusing as they're used to txid referring to the ID of the payment. IMO, this should be changed to txid to be more in line with the existing wallets/explorers out there.

tanx:
Closes #526, closes #527

### DIFF
--- a/src/view/transaction-detail.js
+++ b/src/view/transaction-detail.js
@@ -18,16 +18,9 @@ const TransactionDetailView = ({ store, nav }) => (
   <Background color={color.blackDark}>
     <MainContent style={styles.content}>
       <Modal title="Transaction Details" onClose={() => nav.goTransactions()}>
-        {store.selectedTransaction.confirmationsLabel ? (
-          <DetailField name="Transaction ID">
-            {store.selectedTransaction.id}
-          </DetailField>
-        ) : null}
-        {store.selectedTransaction.hash ? (
-          <DetailField name="Invoice ID">
-            {store.selectedTransaction.hash}
-          </DetailField>
-        ) : null}
+        <DetailField name={store.selectedTransaction.idName}>
+          {store.selectedTransaction.id}
+        </DetailField>
         <DetailField name="Type">
           {store.selectedTransaction.typeLabel}
         </DetailField>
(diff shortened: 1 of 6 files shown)


src/view/transaction-detail.js
```diff
@@ -18,7 +18,16 @@ const TransactionDetailView = ({ store, nav }) => (
...
+        {store.selectedTransaction.confirmationsLabel ? (
+          <DetailField name="Transaction ID">
+            {store.selectedTransaction.id}
+          </DetailField>
+        ) : null}
+        {store.selectedTransaction.hash ? (
+          <DetailField name="Invoice ID">
+            {store.selectedTransaction.hash}
+          </DetailField>
+        ) : null}
```
valentinewallace: Could this be simplified to:
```
        {store.selectedTransaction.confirmationsLabel ? (
          <DetailField name="Transaction ID">
            {store.selectedTransaction.id}
          </DetailField>
        ) : (
          <DetailField name="Invoice ID">
            {store.selectedTransaction.hash}
          </DetailField>
        )}

```
since won't a transaction either be on-chain *or* an invoice?

tanx:
> Looks good! I like this feature especially for being able to easily monitor my transactions on block explorers.

I fixed a bug that was still in the PR. The transaction list only displayed the `id` attribute. So I use that internally for all transaction types and compute the attribute title. Maybe take another look.

> Given that other wallets are using Transaction ID instead of 'Transaction Hash`, I agree with the decision to stay consistent with that. :)

Yeah, all IDs are basically hashes, but only computer scientists know what a hash is.

valentinewallace:
> I fixed a bug that was still in the PR. The transaction list only displayed the id attribute. 

For this screen? I seem to be getting all the fields on commit `c4c1773d0f1067625bc422ed433a20bdf25a5726`.. 
deepinscreenshot_select-area_20180822125604

Just wondering so I can watch out for next time :) 

I can't test with invoices right now because testnet is mining blocks too quickly to sync, but the changes look good to me so I'm good with merging this :)

tanx:
> Just wondering so I can watch out for next time :)

Yeah the bug only applied to invoices since. It's fixed now since all transaction types have an `id` attribute.

PR #543, opened 2018-08-21
Use neutrino in integration tests in stead of RPC.

### LINKED ISSUE #518: Use Neutrino in integration tests instead of RPC

valentinewallace:
Closes #518.

### DIFF
--- a/public/lnd-child-process.js
+++ b/public/lnd-child-process.js
@@ -57,27 +57,38 @@ module.exports.startLndProcess = async function({
 }) {
   if (!lndSettingsDir) throw new Error('lndSettingsDir not set!');
   const processName = 'lnd';
-  const useNeutrino = !isDev && !lndArgs.length;
   let args = [
     '--bitcoin.active',
-    isDev ? '--bitcoin.simnet' : '--bitcoin.testnet',
-    isDev ? '--btcd.rpcuser=kek' : '',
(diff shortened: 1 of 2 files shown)


erkarl:
Running the above changes results in the integration tests hanging after `should get public key node1`.

screen shot 2018-08-21 at 11 06 18

public/lnd-child-process.js
```diff
@@ -57,27 +57,38 @@ module.exports.startLndProcess = async function({
...
+    ]);
+  }
+  // set default production settings if no custom flags
+  if (!isDev && !lndArgs.length) {
+    args = args.concat([
+      '--bitcoin.node=neutrino',
+      '--neutrino.connect=btcd0.lightning.engineering',
+      '--neutrino.connect=127.0.0.1:18333',
+    ]);
+  }
```
tanx: @valentinewallace I refactored the settings to always use Neutrino for integration tests as well as running the app in dev mode. Let me know what you think

valentinewallace: I think that makes sense. That occurred to me as well, but I wasn't sure what we thought our users running in dev mode would want. LGTM, i'll merge! :)

erkarl:
Interesting, when clearing the `data/lnd` folders the integration tests started passing on this branch, too. LGTM 👍

PR #552, opened 2018-08-23
Enable macaroons

valentinewallace:
Due to a security vulnerability, macaroons should always be enabled within the app.

### DIFF
--- a/assets/script/setup_local_cluster.md
+++ b/assets/script/setup_local_cluster.md
@@ -11,19 +11,19 @@ After initiating the wallet and setting your password close the app again.
 ```
 btcd --txindex --simnet --rpcuser=kek --rpcpass=kek --datadir=data/btcd/data --logdir=data/btcd/logs
 
-lnd --rpclisten=localhost:10009 --listen=localhost:10019 --restlisten=localhost:8009 --lnddir=data/lnd --debuglevel=info --bitcoin.simnet --bitcoin.active --bitcoin.node=btcd --btcd.rpcuser=kek --btcd.rpcpass=kek --no-macaroons
+lnd --rpclisten=localhost:10009 --listen=localhost:10019 --restlisten=localhost:8009 --lnddir=data/lnd --debuglevel=info --bitcoin.simnet --bitcoin.active --bitcoin.node=btcd --btcd.rpcuser=kek --btcd.rpcpass=kek
 
-lncli --rpcserver=localhost:10009 --no-macaroons --lnddir=data/lnd unlock
+lncli --rpcserver=localhost:10009 --lnddir=data/lnd unlock
 
-lnd --rpclisten=localhost:10002 --listen=localhost:10012 --restlisten=localhost:8002 --lnddir=data/lnd2 --debuglevel=info --bitcoin.simnet --bitcoin.active --bitcoin.node=btcd --btcd.rpcuser=kek --btcd.rpcpass=kek --no-macaroons --noencryptwallet
+lnd --rpclisten=localhost:10002 --listen=localhost:10012 --restlisten=localhost:8002 --lnddir=data/lnd2 --debuglevel=info --bitcoin.simnet --bitcoin.active --bitcoin.node=btcd --btcd.rpcuser=kek --btcd.rpcpass=kek --noencryptwallet
 ```
 
 ## Fund wallets addresses
 
 ```
-lncli --rpcserver=localhost:10009 --no-macaroons --lnddir=data/lnd newaddress np2wkh
+lncli --rpcserver=localhost:10009 --lnddir=data/lnd newaddress np2wkh
 
-lncli --rpcserver=localhost:10002 --no-macaroons --lnddir=data/lnd2 newaddress np2wkh
+lncli --rpcserver=localhost:10002 --lnddir=data/lnd2 newaddress np2wkh
 
 btcd --txindex --simnet --rpcuser=kek --rpcpass=kek --datadir=data/btcd/data --logdir=data/btcd/logs --miningaddr=NEW_ADDRESS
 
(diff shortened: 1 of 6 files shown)


public/grpc-client.js
```diff
@@ -32,23 +32,19 @@ async function getCredentials(lndSettingsDir) {
...
-function getMetadata(lndSettingsDir) {
-  const metadata = new grpc.Metadata();
-  const macaroonPath = path.join(lndSettingsDir, 'admin.macaroon');
-  const macaroonHex = fs.readFileSync(macaroonPath).toString('hex');
-  metadata.add('macaroon', macaroonHex);
-  return metadata;
+function getMacaroonCreds(lndSettingsDir) {
+  return grpc.credentials.createFromMetadataGenerator(function(args, callback) {
+    const metadata = new grpc.Metadata();
+    const macaroonPath = path.join(lndSettingsDir, 'admin.macaroon');
```
erkarl: It's worth noting that soon the macaroons will be stored in the network directory. Do we want to hold off on this PR until this happens?

test/integration/action/action-integration.spec.js
```diff
@@ -226,6 +221,7 @@ describe('Action Integration Tests', function() {
 
     it('should fund wallet for node1', async () => {
       btcdProcess.kill();
+      await nap(NAP_TIME);
```
erkarl: Flaky test :D?

valentinewallace: Ha, fixed thx

valentinewallace:
Build's failing because we're not using the latest lnd, will update.

public/grpc-client.js
```diff
@@ -72,6 +73,11 @@ module.exports.init = async function({
   });
 
   ipcMain.on('lndInit', event => {
+    const macaroonCreds = getMacaroonCreds(lndSettingsDir, network);
+    credentials = grpc.credentials.combineChannelCredentials(
+      credentials,
+      macaroonCreds
+    );
```
tanx: So the locker doesn't need macaroons? Only the `lnrpc.Lightning` service?

public/grpc-client.js
```diff
@@ -32,23 +32,27 @@ async function getCredentials(lndSettingsDir) {
...
+  return grpc.credentials.createFromMetadataGenerator(function(args, callback) {
+    const metadata = new grpc.Metadata();
+    const macaroonPath = path.join(
+      lndSettingsDir,
+      `data/chain/bitcoin/${network}/admin.macaroon`
+    );
+    const macaroonHex = fs.readFileSync(macaroonPath).toString('hex');
+    metadata.add('macaroon', macaroonHex);
+    callback(null, metadata);
+  });
```
tanx: How does this work with return if it's asynchronous? Don't we need to wait for the callback to set the metadata?

public/grpc-client.js
```diff
@@ -32,23 +32,27 @@ async function getCredentials(lndSettingsDir) {
...
 }
 
-function getMetadata(lndSettingsDir) {
-  const metadata = new grpc.Metadata();
-  const macaroonPath = path.join(lndSettingsDir, 'admin.macaroon');
-  const macaroonHex = fs.readFileSync(macaroonPath).toString('hex');
-  metadata.add('macaroon', macaroonHex);
-  return metadata;
+function getMacaroonCreds(lndSettingsDir, network) {
+  return grpc.credentials.createFromMetadataGenerator(function(args, callback) {
```
tanx: Use array function `(args, callback) =>` here.

public/grpc-client.js
```diff
@@ -72,6 +73,11 @@ module.exports.init = async function({
   });
 
   ipcMain.on('lndInit', event => {
+    const macaroonCreds = getMacaroonCreds(lndSettingsDir, network);
+    credentials = grpc.credentials.combineChannelCredentials(
+      credentials,
+      macaroonCreds
+    );
     lnd = new lnrpc.Lightning(`localhost:${lndPort}`, credentials);
```
tanx: So we don't need to pass the metadata in each api call because we inject the credentials here?

src/config.js
```diff
@@ -9,7 +9,7 @@ module.exports.RATE_DELAY = 15 * 60 * 1000;
 
 module.exports.LND_PORT = 10009;
 module.exports.LND_PEER_PORT = 10019;
-module.exports.MACAROONS_ENABLED = false;
+module.exports.NETWORK = 'testnet';
```
tanx: What if we're in dev mode? Then the app will start with `testnet` instead of `simnet`?

assets/script/setup_local_cluster.md
```diff
@@ -33,19 +33,19 @@ btcctl --simnet --rpcuser=kek --rpcpass=kek generate 400
...
 btcctl --simnet --rpcuser=kek --rpcpass=kek generate 6
 
-lncli --rpcserver=localhost:10009 --no-macaroons --lnddir=data/lnd listchannels
+lncli --rpcserver=localhost:10009 --lnddir=data/lnd listchannels
 
-lncli --rpcserver=localhost:10002 --no-macaroons --lnddir=data/lnd2 addinvoice --amt=10000
+lncli --rpcserver=localhost:10002 --lnddir=data/lnd2 addinvoice --amt=10000
 
-lncli --rpcserver=localhost:10009 --no-macaroons --lnddir=data/lnd sendpayment --pay_req=ENCODED_INVOICE
+lncli --rpcserver=localhost:10009 --lnddir=data/lnd sendpayment --pay_req=ENCODED_INVOICE
```
tanx: Nice update to the docs 👍

public/grpc-client.js
valentinewallace: Yep, thought it simplified the code a bit to not have to remember to pass in `metadata` each time.

public/grpc-client.js
valentinewallace: If the wallet hasn't been initialized, the macaroons don't exist yet. The wallet creation process involves creating the macaroons :)

public/grpc-client.js
valentinewallace: If we stuck with passing in metadata each time, we'd also have to get the macaroons to `action/grpc` and pass them into the `Duplex` here: https://github.com/lightninglabs/lightning-app/blob/master/src/action/grpc.js#L96

tanx: Ok great 👍

public/electron.js
```diff
@@ -123,7 +123,7 @@ function createWindow() {
     ipcMain,
     lndSettingsDir,
     lndPort: LND_PORT,
-    macaroonsEnabled: MACAROONS_ENABLED,
+    network: NETWORK,
```
tanx: `network: isDev ? 'simnet' : NETWORK`
Else the app will use `'testnet'` in dev mode as well.

public/grpc-client.js
valentinewallace: In order to combine the ssl credentials with the macaroon metadata, we need to turn the metadata into a `CallCredentials` object. 

`createFromMetadataGenerator` requires an metadata generator (the asynchronous function) and returns a `CallCredentials` object that can be composed with the ssl `ChannelCredentials` object: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1346#issuecomment-240912476

tanx: Ok. Thanks for the explanation.

PR #564, opened 2018-08-28
Switch from killing lnd process on app close to using StopDaemon RPC …

valentinewallace:
…command

This prevents loss of user data by ensuring safe shutdown of the daemon.

This functionality doesn't work perfectly and improved functionality is waiting on a few changes to be merged into `lnd`. Users will need to use `ctrl-c` to quit the app if the RPC call is not available at the time of app close.

### DIFF
--- a/public/electron.js
+++ b/public/electron.js
@@ -38,7 +38,6 @@ const lndArgs = process.argv.filter(a =>
 // Keep a global reference of the window object, if you don't, the window will
 // be closed automatically when the JavaScript object is garbage collected.
 let win;
-let lndProcess;
 let btcdProcess;
 
 log.transports.console.level = 'info';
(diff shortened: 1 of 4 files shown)


erkarl:
Tested this out locally and I'm unable to get `[INF] LTND: Shutdown complete` message in the logs. I'm also seeing the `[error] Caught Main Process Error:` errors - not sure if they are related.

I tried closing the app multiple times by:
1. cmd + x
2. just closing the window by clicking the X button

cfromknecht:
Since we're spawning the child process locally, we may want to consider using SIGINT and bypass the RPC altogether. I know @erkarl reported that the results were inconsistent in testing, were you able to get more reliable results?

cfromknecht:
Stopdaemon is kind of nice that it is agnostic to whether lnd is running locally or not, though in the case that we're connecting remote we may not want to call `StopDaemon`?

PR #571, opened 2018-08-29
Wait for filter headers

### LINKED ISSUE #515: Autopilot does not open channels when using Neutrino
It seems to work when running a local full node. But I still have to restart several times.

tanx:
Closes #515


src/store.js
```diff
@@ -25,9 +25,11 @@ export class Store {
       walletUnlocked: false, // Is the wallet unlocked
       lndReady: false, // Is lnd process running
       syncedToChain: false, // Is lnd synced to blockchain
+      isSyncing: false, // Is lnd syncing to chain or filter headers
```
erkarl: `syncedToChain` property 1 line above is only being used in `initLoaderSyncing` action. Should we use `isSyncing` there as well?

In addition the naming of these 2 is confusing. I do understand what the difference is by reading the comment, but it won't be apparent in other places in the code.

erkarl:
@tanx the changes to naming look good.

I'm having trouble getting the headers to sync within simnet when testing manually. What are the steps that I should take?

tanx:
> I'm having trouble getting the headers to sync within simnet when testing manually. What are the steps that I should take?

Just tested simnet. Starting with an existing data dir works from me. I think @valentinewallace was also having issues on simnet. Are you guys setting up the app from scratch?

erkarl:
@tanx my workflow was to delete everything in the data dir and then run `electron-dev`.

tanx:
> @tanx my workflow was to delete everything in the data dir and then run electron-dev.

Did you follow the simnet cluster steps to connect to another lnd peer?

erkarl:
@tanx Yep. I'll test again one more time tomorrow morning just to make sure.

PR #575, opened 2018-08-29
Node libs

### LINKED ISSUE #537: Setup react storybook for mobile
See: https://github.com/storybooks/storybook/tree/master/app/react-native

tanx:
Closes #537

### DIFF
--- a/mobile/App.js
+++ b/mobile/App.js
@@ -1,2 +1,3 @@
+import 'node-libs-react-native/globals';
 import Storybook from './storybook';
 export default Storybook;
--- a/mobile/rn-cli.config.js
+++ b/mobile/rn-cli.config.js
@@ -11,6 +11,10 @@ module.exports = {
     ),
     mobx: path.resolve(__dirname, 'node_modules/mobx'),
     '@storybook': path.resolve(__dirname, 'node_modules/@storybook'),
+    'node-libs-react-native': path.resolve(
+      __dirname,
+      'node_modules/node-libs-react-native'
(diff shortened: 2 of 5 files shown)


mobile/rn-cli.config.js
```diff
@@ -10,7 +10,13 @@ module.exports = {
       'node_modules/react-native-svg'
     ),
     mobx: path.resolve(__dirname, 'node_modules/mobx'),
+    'mobx-react': path.resolve(__dirname, 'node_modules/mobx-react'),
```
valentinewallace: According to [this](https://gist.github.com/parshap/e3063d9bf6058041b34b26b7166fd6bd#global-environment) gist, if we're using the `rn-cli.config.js` solution to include node modules in react native, we'll also need to set the React Native global environment because "node-style" requires expect the global environment to be in a specific state.

mobile/App.js
```diff
@@ -1,2 +1,3 @@
+import 'node-libs-react-native/globals';
```
tanx: @valentinewallace the globals like `Buffer` and `process` are set here.

mobile/rn-cli.config.js
tanx: See comment above...

erkarl:
I noticed we added `sinon` dependency here to mobile, but we're actually not using it. I'm assuming this is infrastructure groundwork for mobile?

This was a pleasure to review commit-by-commit 👍 . LGTM.

PR #620, opened 2018-09-06
Fix autopilot + syncing speed.

valentinewallace:
By compiling with experimental lnd and setting --routing.assumechanvalid, autopilot is enabled and will start creating channels even before cfheaders are synced. Additionally, syncing should now take <10 minutes rather than over an hour.

### DIFF
--- a/assets/script/install_lnd.sh
+++ b/assets/script/install_lnd.sh
@@ -2,7 +2,7 @@
 
 # versions
 GO_TAG=1.10.3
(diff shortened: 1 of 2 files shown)


tanx:
@Roasbeef does this set `synced_to_chain` to true? IT keep polling for me
bildschirmfoto 2018-09-06 um 07 43 01

halseth:
@tanx Can you check debug logs?

Issue #631, opened 2018-09-07
Windows build is stuck at the welcome screen


tanx:

erkarl:
Was able to replicate the issue on a fresh windows 10 installation (within a VM):
![screen shot 2018-09-07 at 13 29 42](https://user-images.githubusercontent.com/1056569/45214083-292c1100-b2a2-11e8-92c6-d23fef636860.png)

tanx:
Ok thanks @erkarl. Can you paste the logs here? Does it also do this in dev mode?

erkarl:
![screen shot 2018-09-07 at 15 00 05](https://user-images.githubusercontent.com/1056569/45218013-c9883280-b2ae-11e8-8eea-0e1cd7b78486.png)

@tanx turns out it's an easy fix. Looks like windows binary was built without the experimental flags turned on as per commit: `https://github.com/lightninglabs/lightning-app/commit/4140e5a7b9f27752937e8fe703a2c10996e1031e#diff-75787c1f3575f24b70d14aac4eaf85eb`

I managed to replicate the issue on OSX as well when I use LND without the experimental flags :).

tanx:
> @tanx turns out it's an easy fix. Looks like windows binary was built without the experimental flags turned on as per commit: 4140e5a#diff-75787c1f3575f24b70d14aac4eaf85eb

Nice! Looks like we need to call the flags in `build_app.sh` as well...

tanx:
See: https://github.com/lightninglabs/lightning-app/pull/633

tanx:
Confirmed it worked for moli

PR #632, opened 2018-09-07
Fix windows binary build

erkarl:
Adds "experimental" tag to the windows binary build so that it
recognizes the `routing.assumechanvalid` flag.

After building the binary with the updated script my windows 10 test machine is now nicely syncing to the chain.

Closes https://github.com/lightninglabs/lightning-app/issues/631


tanx:
utACK awesome! thanks @erkarl 

I'll trigger a new build so we can reproduce the fix in the release 👍

PR #633, opened 2018-09-07
Revert "Fix windows binary build"

tanx:
Reverts lightninglabs/lightning-app#632

@erkarl did you test the script? Seems like this broke the Windows build

bildschirmfoto 2018-09-07 um 17 27 23


erkarl:
@tanx Sorry for that.
I did test the script locally and then ran the binary with success on windows. What was the error message?

PR #635, opened 2018-09-08
Use different ports for lnd/grpc in case user already running lnd

### LINKED ISSUE #630: Use different ports for lnd/grpc in case a user already has an lnd node running on localhost

valentinewallace:
Closes #630.


erkarl:
Tested with another LND instance running on port `10009` - seems to work fine. 👍 

Should we also update the ports in https://github.com/lightninglabs/lightning-app/blob/master/assets/script/setup_local_cluster.md for consistency?

tanx:
LGMT

> Should we also update the ports in https://github.com/lightninglabs/lightning-app/blob/master/assets/script/setup_local_cluster.md for consistency?

Done

Issue #544, opened 2018-08-21
Transaction list with lots of entries is slow


erkarl:
### Description
Transaction list with lots of entries is very janky.

### Steps to reproduce the behavior
Using simnet mine to block 2400 so that the transaction list has many transaction list items.

### Expected behavior
Transaction list opens immediately and incrementally loads more entries when there's enough room in the performance budget. The maximum render time should be 1000 (second) / 60 (frames per second) = ~16.67ms so that there's no noticeable slow down for the end user.

### Actual behavior
On a powerful desktop CPU (i7-7700k) clicking on the transactions results in a noticeable delay - making the app feel slow.

### Screenshot and/or logs if applicable
![screen shot 2018-08-21 at 10 45 21](https://user-images.githubusercontent.com/1056569/44387852-556c3180-a52f-11e8-9cdc-b82aedbb5ed1.png)

tanx:
@erkarl thanks for benchmarking. Paging transactions for lnd is in the pipeline. In the meantime we could mitigate the performance issue by calling `return all.slice(0, 100)` when computing `store.computedTransactions`.

PR #641, opened 2018-09-10
Limit transactions list to last 100 transactions

erkarl:
Closes https://github.com/lightninglabs/lightning-app/issues/544

This PR mitigates the performance issue when there are lots of computed transactions in the transactions list.

screen shot 2018-09-10 at 14 15 07

The app still feels a little janky/sluggish (~800ms frame), but it's definitely an improvement and quick win over the previous result (~2000ms).

### DIFF
--- a/src/computed/transaction.js
+++ b/src/computed/transaction.js
@@ -28,7 +28,7 @@ const ComputedTransaction = store => {
           t.confirmationsLabel = t.confirmations.toString();
         }
       });
-      return all;
+      return all.slice(0, 100);
     }),
   });
 };
--- a/test/unit/computed/transaction.spec.js
+++ b/test/unit/computed/transaction.spec.js
@@ -3,18 +3,19 @@ import ComputedTransaction from '../../../src/computed/transaction';
 
 describe('Computed Transactions Unit Tests', () => {
   let store;
+  const bitcoinTransaction = {
+    id: '0',
+    type: 'bitcoin',
+    amount: 923456,
+    fee: 8250,
+    confirmations: 0,
+    status: 'unconfirmed',
+    date: new Date(),
+  };
 
   beforeEach(() => {
     store = new Store();
-    store.transactions.push({
-      id: '0',
-      type: 'bitcoin',
-      amount: 923456,
-      fee: 8250,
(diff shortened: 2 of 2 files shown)


src/computed/transaction.js
```diff
@@ -12,8 +12,9 @@ const ComputedTransaction = store => {
       const t = transactions ? transactions.slice() : [];
       const p = payments ? payments.slice() : [];
       const i = invoices ? invoices.slice() : [];
-      const all = [].concat(t, p, i);
+      let all = [].concat(t, p, i);
       all.sort((a, b) => b.date.getTime() - a.date.getTime());
+      all = all.slice(0, 100);
```
tanx: It's fine to do the slice in the return value (line 31). The performance bottleneck isn't V8 but rather the DOM.

test/unit/computed/transaction.spec.js
```diff
@@ -68,5 +68,43 @@ describe('Computed Transactions Unit Tests', () => {
...
+          transaction.status = randomNumber({ min: 0, max: 1 })
+            ? 'complete'
+            : 'in-progress';
+          if (transactionType === 3) {
+            transaction.fee = randomNumber({ min: 0, max: 250 });
+            store.payments.push(transaction);
+          } else {
+            store.invoices.push(transaction);
+          }
+        }
```
tanx: Probably easier to just clone `store.transactions` 101 times. No need for all the test logic here...

test/unit/computed/transaction.spec.js
```diff
@@ -68,5 +68,43 @@ describe('Computed Transactions Unit Tests', () => {
...
+      });
+      ComputedTransaction(store);
+      expect(store.computedTransactions.length, 'to equal', 100);
```
tanx: ... since all we need to check is this.

erkarl:
@tanx thanks for the feedback - updated.

tanx:
> The app still feels a little janky/sluggish (~800ms frame), but it's definitely an improvement and quick win over the previous result (~2000ms).

Ok. Yeah a good first step. We should look into how we can improve performance further with fancier optimizations later.

Issue #642, opened 2018-09-10
Use font with equally wide number characters for balance label in Payment Request View


wallacewest:
### Description

Currently used WorkSans does not have equally wide number characters, which makes it harder to create input field with adjustable font size. A less performant or less maintainable solution is thus required.

### Steps to reproduce the behavior

Visit `Payment Request Screen`

### Expected behavior

A traditional font, where every number character takes exactly the same width (OpenSans), should be used. A downside of such solution is that it will usually take more horizontal space for the same string of numbers.

### Actual behavior

When you calculate font size based on number of characters and available width, you are left with some white space.

### Needed to resolve issue No.

#294

wallacewest:
Made obsolete by PR #643

Issue #615, opened 2018-09-04
AmountInputField behaviour not constrained


joostjager:
### Description

Various undesired states can be created in the amount input field.

![image](https://user-images.githubusercontent.com/4638168/45038848-6609c500-b062-11e8-9686-6797df823735.png)

- Negative amount
- Multiple leading zeros
- More than two decimals
- Label gets too large

- Entering `0-`

![image](https://user-images.githubusercontent.com/4638168/45038914-95203680-b062-11e8-9fb5-0f5e5f93660e.png)

- Multiple dots `...4`

![image](https://user-images.githubusercontent.com/4638168/45038949-ae28e780-b062-11e8-9e44-01a87b741244.png)

### Steps to reproduce the behavior

Using `Lightning-linux-x86_64v0.2.0-prealpha.25.AppImage`

- Go to payment request screen
- Enter invalid values

### Expected behavior

At minimum, any value that is not a valid amount should be impossible to enter (negative, three decimals). 

In addition to that, it would be good to keep the formatting consistent. So either `.50` or `0.50` but not make both possible. Same for `3.5` vs `3.50`.

### Actual behavior

See screenshots

valentinewallace:
@tanx If you have PayPal, for fiat input I'm making it behave exactly like the PayPal's input for making a payment. If you could try it out/sign off on that heuristic, that would be great since it's an opinionated UX :)

tanx:
@valentinewallace do you mean the iOS app for paypal? That looks good to me, but it only has a number pad as keyboard.

The problem is that we'd need to create a custom React Native `TextInput` since we're just using that standard component here. Perhaps there are some flags/options we can set to prevent negative numbers and multiple leading zeros. Although it doesn't seem too big a deal IMHO. lnd should throw an error for invalid amounts anyway if the user tries to submit.

PR #643, opened 2018-09-10
Resize balance label font size based on content width in Payment Request View

### LINKED ISSUE #294: Resize balance label font size based content width in Payment Request View
In order for the Balance number to resize correctly we should add support to react-native-web for `adjustsFontSizeToFit`. I created an issue upstream: https://github.com/necolas/react-native-web/issues/889

wallacewest:
Closes #294 and #615 (except the decimals limit and optional leading zero, for I don't see anything being wrong with these).

Makes the issue #642 obsolete.


wallacewest:
Once you enter ~25+ characters on the Electron version of the app, the rounding of computed font size stars becoming an issue. =/

PR #667, opened 2018-09-12
Payment timeout

### LINKED ISSUE #555: Set a 1 minute timeout when sending a lightning payment and display "no route found" screen
If no route is found, payments can take forever and the wait spinner just sits there and the app is unresponsive.

We could set a timeout before calling the `sendPayment` grpc api in the `payLightning` payment action. If the timeout is reached, we can display the "no route found" view.

This approach would make the user experience more responsive. The downside is if the user retries the payment a second time and the first payment does eventually go through in the background, he will have paid twice.

valentinewallace:
Time out payments after 1 minute. 

Closes #555

### DIFF
--- a/src/view/main.js
+++ b/src/view/main.js
@@ -113,7 +113,7 @@ class MainView extends Component {
           <PayLightningDone store={store} payment={payment} nav={nav} />
         )}
         {route === 'PaymentFailed' && (
-          <PaymentFailed channel={channel} payment={payment} />
+          <PaymentFailed channel={channel} nav={nav} />
         )}
         {route === 'PayBitcoin' && (
           <PayBitcoin store={store} payment={payment} nav={nav} />
--- a/src/view/payment-failed.js
+++ b/src/view/payment-failed.js
@@ -28,7 +28,7 @@ const styles = StyleSheet.create({
   },
 });
 
-const PaymentFailedView = ({ channel, payment }) => (
+const PaymentFailedView = ({ channel, nav }) => (
   <Background color={color.blackDark}>
     <MainContent>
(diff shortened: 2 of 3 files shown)


src/action/payment.js
```diff
@@ -174,17 +174,26 @@ class PaymentAction {
       this._nav.goWait();
       const invoice = this._store.payment.address.replace(PREFIX_URI, '');
       const stream = this._grpc.sendStreamCommand('sendPayment');
-      await new Promise((resolve, reject) => {
+      const success = await new Promise((resolve, reject) => {
+        const timeout = setTimeout(() => resolve(false), PAYMENT_TIMEOUT);
         stream.on('data', data => {
           if (data.payment_error) {
+            clearTimeout(timeout);
```
erkarl: Looks like we're clearing the timeout here regardless of the conditional. Can we move it outside the conditional (below line 179) and delete line 184?

src/action/payment.js
```diff
@@ -170,6 +170,11 @@ class PaymentAction {
    * @return {Promise<undefined>}
    */
   async payLightning() {
+    let failed = false;
+    const timeout = setTimeout(() => {
```
erkarl: How can I test this block manually? In my local simnet environment I'm always getting an error from the stream `data.payment_error` and it just shows the notification.

valentinewallace: I tested by manually putting a sleep before the failed check, forcing a timeout. Not the most robust :grimacing:

erkarl: @valentinewallace well, that's one way of doing it :D.

PR #676, opened 2018-09-13
Set invoice as not private to add routing hint

### LINKED ISSUE #675: Add routing hints in invoice
We can do this by setting `private` to true in https://api.lightning.community/#addinvoice

tanx:
Closes #675

### DIFF
--- a/src/action/invoice.js
+++ b/src/action/invoice.js
@@ -52,6 +52,8 @@ class InvoiceAction {
    * which can be rendered in a QR code for scanning. After the values
    * are set on the store the user is navigated to the invoice QR view
    * which displays the QR for consumption by the payer.
+   * The invoice is set to not private since it should contain routing
+   * information for private channels.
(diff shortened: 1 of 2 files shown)


src/action/invoice.js
```diff
@@ -62,6 +64,7 @@ class InvoiceAction {
       const response = await this._grpc.sendCommand('addInvoice', {
         value: toSatoshis(invoice.amount, settings),
         memo: invoice.note,
+        private: false,
```
erkarl: @tanx I think it's the opposite. If it's `private: true` it will encode routing hints in the invoice with private channels in order to assist the payer in reaching you. 

During my simnet test I wasn't able to pay the the invoice from `LND3 --non-private--> LND2 --private--> LND1`. When changing it to `true` I was able to generate an invoice from the app and pay it.

tanx: @erkarl good catch! I fixed it.

PR #679, opened 2018-09-13
Fix network timer

### LINKED ISSUE #668: Improve loader - syncing chain spinner
Currently, the spinner will get to 100% when the backend *begins* syncing the filter headers. Therefore, it will stay at 100% for up to 10 or so minutes.

valentinewallace:
Closes #668.


valentinewallace:
Hm, I didn't realize the lnd side fix was that imminent. If that's the case, I'd rather wait on that.

PR #680, opened 2018-09-14
Do not display fees under transaction details when fees are empty

erkarl:
Closes https://github.com/lightninglabs/lightning-app/issues/673

Not sure if/how we should test this. If do not assign fees to `-` under `computedTransactions` then the transaction list will look weird with non-empty cells.


valentinewallace:
Couldn't it be potentially confusing to omit the fee field, and it would be better to just list it as `0 BTC` in this situation? 

Can't see why a user would assume 0 fee in the case where the field is missing altogether.

erkarl:
@valentinewallace yep, it could be confusing for users. What about displaying `0` instead of `-`?

valentinewallace:
@erkarl I think that makes more sense. 

@tanx Let us know what you think.

tanx:
Agreed on displaying `0` being better.

erkarl:
Updated.

PR #687, opened 2018-09-17
Enable force close channel

erkarl:
Closes: https://github.com/lightninglabs/lightning-app/issues/481

### DIFF
--- a/src/action/channel.js
+++ b/src/action/channel.js
@@ -259,11 +259,14 @@ class ChannelAction {
    * the necessary error handling and notification display.
    * @return {Promise<undefined>}
    */
-  async closeSelectedChannel() {
+  async closeSelectedChannel(force = false) {
     try {
       const { selectedChannel } = this._store;
       this._nav.goChannels();
-      await this.closeChannel({ channelPoint: selectedChannel.channelPoint });
+      await this.closeChannel({
+        channelPoint: selectedChannel.channelPoint,
+        force,
+      });
     } catch (err) {
       if (
         err &&
--- a/src/action/channel.js
+++ b/src/action/channel.js
@@ -265,7 +265,15 @@ class ChannelAction {
       this._nav.goChannels();
       await this.closeChannel({ channelPoint: selectedChannel.channelPoint });
     } catch (err) {
-      this._notification.display({ msg: 'Closing channel failed!', err });
+      if (
+        err &&
+        err.details &&
+        err.details.includes('try force closing it instead')
+      ) {
+        this._nav.goChannelForceDelete();
+      } else {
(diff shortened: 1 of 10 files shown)


src/action/channel.js
```diff
@@ -259,13 +259,24 @@ class ChannelAction {
...
+      await this.closeChannel({
+        channelPoint: selectedChannel.channelPoint,
+        force,
+      });
     } catch (err) {
-      this._notification.display({ msg: 'Closing channel failed!', err });
+      if (
+        err &&
+        err.details &&
+        err.details.includes('try force closing it instead')
```
erkarl: We're checking the string contents of the error message here because LND returns a generic/unknown error code.

https://github.com/lightningnetwork/lnd/blob/master/rpcserver.go#L1138
https://github.com/grpc/grpc-go/blob/master/codes/codes.go#L43

valentinewallace:
@erkarl I'm going to talk to our designer first, because I'm not sure exactly how this feature is going to look currently. We want to warn the user somehow of the tradeoffs.

erkarl:
@valentinewallace sounds good. With the current implementation I used the same "close channel" screen.

https://github.com/lightninglabs/lightning-app/pull/687/files#diff-95d6f755680ac8d4f60b8b3b91955aefR33

valentinewallace:
@erkarl So it turns out the desired behavior for these screens is as follows:

For [this screen](https://invis.io/UHN289CASW9#/281222453_Channels_Detail):
If cooperative close is not available for whatever reason, say "force close channel" at bottom instead of "close channel".

For [this screen](https://invis.io/UHN289CASW9#/282808116_Channel_-_Confirm_Delete):
If the channel is cooperative closing, show the copy shown on the screen.
If the channel needs to be forced closed, change the copy to warn the user how long their funds will be unavailable/potentially other tradeoffs of force closing (I think that's the only one though). 

These are pretty significant changes to how this PR currently is, so let me know if you're still interested.

erkarl:
@valentinewallace thanks for the update. 

If I'm not mistaken the channel delete view already works as you described:
1. User enters the `close-channel` view and clicks "Close Channel"
2. If the channel close is unsuccessful a copy of `channel-force-delete` will be shown, explaining the trade-offs (except CSV delay)

To conclude this PR needs to be updated with:
1. Change "Close Channel" to "Force Close Channel" if the channel is inactive in [this screen](https://projects.invisionapp.com/share/UHN289CASW9#/281222453_Channels_Detail)
2. Include CSV delay in the `force-close-channel` view. Or perhaps we don't need a separate view for this and just change the copy text and make the close channel/force close channel button behaviour dynamic within the `channel-delete` view?

tanx:
Closing in favor of #717

Issue #628, opened 2018-09-07
Clear password field if incorrectly entered


valentinewallace:
Currently, if a user mistypes their password the field won't clear on submit.

PR #698, opened 2018-09-20
Clear password field if incorrectly entered

erkarl:
Closes: https://github.com/lightninglabs/lightning-app/issues/628

### DIFF
--- a/test/unit/action/wallet.spec.js
+++ b/test/unit/action/wallet.spec.js
@@ -355,16 +355,15 @@ describe('Action Wallet Unit Tests', () => {
     });
 
     it('should display error notification on failure and clear password', async () => {
-      sandbox.stub(wallet, 'setPassword');
+      wallet.setPassword('not-empty');
+      expect(store.wallet.password, 'not to be', '');
       grpc.sendUnlockerCommand
         .withArgs('UnlockWallet')
         .rejects(new Error('Boom!'));
       await wallet.unlockWallet({ walletPassword: 'baz' });
       expect(notification.display, 'was called once');
       expect(nav.goWait, 'was not called');
-      expect(wallet.setPassword, 'was called with', {
-        password: '',
-      });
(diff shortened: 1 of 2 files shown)


test/unit/action/wallet.spec.js
```diff
@@ -354,13 +354,17 @@ describe('Action Wallet Unit Tests', () => {
...
+      sandbox.stub(wallet, 'setPassword');
       grpc.sendUnlockerCommand
         .withArgs('UnlockWallet')
         .rejects(new Error('Boom!'));
       await wallet.unlockWallet({ walletPassword: 'baz' });
       expect(notification.display, 'was called once');
       expect(nav.goWait, 'was not called');
+      expect(wallet.setPassword, 'was called with', {
+        password: '',
+      });
```
tanx: I wouldn't stub `setPassword` in this test since that's the logic you wanna test. Instead, I'd check if `store.wallet.password` equals `''`.

test/unit/action/wallet.spec.js
```diff
@@ -354,13 +354,16 @@ describe('Action Wallet Unit Tests', () => {
...
+    it('should display error notification on failure and clear password', async () => {
+      wallet.setPassword('not-empty');
+      expect(store.wallet.password, 'not to be', '');
       grpc.sendUnlockerCommand
         .withArgs('UnlockWallet')
         .rejects(new Error('Boom!'));
       await wallet.unlockWallet({ walletPassword: 'baz' });
       expect(notification.display, 'was called once');
       expect(nav.goWait, 'was not called');
+      expect(store.wallet.password, 'to be', '');
```
erkarl: @tanx I've updated the PR according the to feedback, but I want to point out that the logic for `setPassword` is already tested in a separate unit test.

https://github.com/lightninglabs/lightning-app/blob/master/test/unit/action/wallet.spec.js#L82-L87

Issue #627, opened 2018-09-07
Don't attempt off-chain payment without the chain being synced


valentinewallace:
Currently, we will attempt an off-chain payment without the chain being synced, which doesn't work. 
Show a notification warning the user when they try to send a payment under these circumstances.

PR #699, opened 2018-09-20
Disable Lightning payments when syncing

erkarl:
Closes: https://github.com/lightninglabs/lightning-app/issues/627

screen shot 2018-09-20 at 15 37 57

### DIFF
--- a/src/action/payment.js
+++ b/src/action/payment.js
@@ -87,7 +87,11 @@ class PaymentAction {
       return this._notification.display({ msg: 'Enter an invoice or address' });
     }
     if (await this.decodeInvoice({ invoice: this._store.payment.address })) {
-      this._nav.goPayLightningConfirm();
+      this._store.syncedToChain
+        ? this._nav.goPayLightningConfirm()
+        : this._notification.display({
+            msg: 'Unable to make Lightning payment when syncing to chain',
+          });
     } else if (isAddress(this._store.payment.address)) {
       this._nav.goPayBitcoin();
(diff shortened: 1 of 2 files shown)


src/action/payment.js
```diff
@@ -87,7 +87,11 @@ class PaymentAction {
       return this._notification.display({ msg: 'Enter an invoice or address' });
     }
     if (await this.decodeInvoice({ invoice: this._store.payment.address })) {
-      this._nav.goPayLightningConfirm();
+      this._store.syncedToChain
+        ? this._nav.goPayLightningConfirm()
+        : this._notification.display({
+            msg: 'Unable to make Lightning payment when syncing to chain',
+          });
```
tanx: It would probably be more readable to put `if (!this._store.syncedToChain)` in its own block and display the notification like in the address check above.

erkarl: Just to clarify... are you suggesting to put the `if (!this._store.syncedToChain)` above the `if (await this.decodeInvoice({ invoice: this._store.payment.address }))`? If that's the case then we'd also be restricting on-chain bitcoin payments until synced.

Alternatively we could replace the ternary operator with:
```
if (this._store.syncedToChain) {
  this._nav.goPayLightningConfirm();
} else {
  this._notification.display({
    msg: 'Unable to make Lightning payment when syncing to chain',
  });
}
```
which doesn't read much better 🤔

erkarl:
Closing due to inactivity.

Issue #611, opened 2018-08-30
Keep column titles at the top of the page on scroll down


valentinewallace:
Currently, if a user has a lot of channels/transactions and they scroll down on the page, the column titles disappear. Instead, the titles should stay at the top of the page while the list of channels scrolls underneath. 

before scroll:
![deepinscreenshot_select-area_20180905124356](https://user-images.githubusercontent.com/10647624/45117042-6a3ced80-b109-11e8-994d-c14eab8e6cd5.png)

after scroll:
![deepinscreenshot_select-area_20180905124408](https://user-images.githubusercontent.com/10647624/45117057-745eec00-b109-11e8-9009-0a098e63f07a.png)

tanx:
Can you post a screenshot?

PR #700, opened 2018-09-21
Make list headers sticky

erkarl:
In this PR we:
* make `transactions`, `channels` and `notifications` list headers sticky
* upgrade `react-native-web` and its dependencies (`react-art`, `react-dom`, `react-dom`)

The upgrade for `react-native-web` was required because previously `stickyHeaderIndices` for `ListView` was buggy. It's worth noting that the new version also supports [FlatList](https://facebook.github.io/react-native/docs/flatlist) and [SectionList](https://facebook.github.io/react-native/docs/sectionlist) which have lots of improvements over the `ListView`.

During the upgrade process, I verified that all the screens and components didn't visually change/break. The only change needed was for the `Seed` view because `flexGrow` property used in `MainContent` component now works according to the spec - it's assigning `flexBasis` to `auto`. I've included the necessary fix for `Seed` view.

In the future we could add visual regression tests to the storybook stories to make sure the views don't change when upgrading libraries etc.

Channels:
channels

Transactions:
transactions

Notifications:
notifications

Closes: https://github.com/lightninglabs/lightning-app/issues/611

### DIFF
--- a/src/component/list.js
+++ b/src/component/list.js
@@ -50,6 +50,7 @@ export class List extends Component {
         ref={component => (this.list = component)}
         dataSource={this.dataSource}
         renderHeader={this.props.renderHeader}
+        stickyHeaderIndices={this.props.stickyHeaderIndices}
         renderRow={this.props.renderItem}
         enableEmptySections={true}
       />
@@ -61,6 +62,7 @@ List.propTypes = {
   data: PropTypes.array.isRequired,
   renderHeader: PropTypes.func,
   renderItem: PropTypes.func.isRequired,
+  stickyHeaderIndices: PropTypes.array,
 };
 
 //
--- a/src/component/list.js
+++ b/src/component/list.js
@@ -50,7 +50,7 @@ export class List extends Component {
         ref={component => (this.list = component)}
         dataSource={this.dataSource}
         renderHeader={this.props.renderHeader}
-        stickyHeaderIndices={this.props.stickyHeaderIndices}
+        stickyHeaderIndices={[0]}
         renderRow={this.props.renderItem}
         enableEmptySections={true}
       />
@@ -62,7 +62,6 @@ List.propTypes = {
   data: PropTypes.array.isRequired,
   renderHeader: PropTypes.func,
   renderItem: PropTypes.func.isRequired,
-  stickyHeaderIndices: PropTypes.array,
 };
 
 //
--- a/src/view/channel.js
+++ b/src/view/channel.js
@@ -107,7 +107,6 @@ const ChannelList = ({ store, channel }) => {
       />
       <ListContent>
         <List
-          stickyHeaderIndices={[0]}
           data={channels}
           renderHeader={ChannelListHeader}
           renderItem={item => (
--- a/src/view/channel.js
+++ b/src/view/channel.js
@@ -107,8 +107,9 @@ const ChannelList = ({ store, channel }) => {
       />
       <ListContent>
         <List
+          stickyHeaderIndices={[0]}
           data={channels}
-          renderHeader={() => <ChannelListHeader />}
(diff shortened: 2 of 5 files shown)


src/component/list.js
```diff
@@ -61,6 +62,7 @@ List.propTypes = {
   data: PropTypes.array.isRequired,
   renderHeader: PropTypes.func,
   renderItem: PropTypes.func.isRequired,
+  stickyHeaderIndices: PropTypes.array,
```
erkarl: `stickyHeaderIndices` was added as a prop because I assumed we might not want to have sticky headers on all lists. If that's not the case we can hardcode `stickyHeaderIndices={[0]}` instead of line 53.

valentinewallace: IMO we do always want sticky headers :)

package.json
```diff
@@ -36,10 +36,10 @@
     "mobx": "^3.6.2",
     "mobx-react": "^4.4.3",
     "qr-image": "^3.2.0",
-    "react": "~16.3.2",
-    "react-art": "~16.3.2",
-    "react-dom": "~16.3.2",
-    "react-native-web": "^0.6.0",
+    "react": "^16.5.2",
```
valentinewallace: Don't think these changes should be part of this PR?

src/component/list.js
erkarl: Cool, I'll make the appropriate changes.

package.json
erkarl: Because `stickyHeaderIndices` for `ListView` were buggy I had to upgrade `react-native-web` and its dependencies `react`, `react-art`, `react-dom`. Should we make the upgrade in a separate PR and rebase this one once the upgrade PR is merged?

tanx: > Don't think these changes should be part of this PR?

This is tricky. I've been putting off upgrading react/react-native-web since there was quite a bit of breakage in some views. @erkarl have you tested all screens in the storybook to make sure there are no regressions? I remember ListViews not scrolling without a manual window resize after items were added dynamically by actions.

erkarl: Good point - I'll make sure to test this when I make the other requested changes in this PR.

erkarl: @tanx I tested the ListView scrolling by dynamically adding items - didn't see any issues.

As for testing the screens for visual regressions - I did test that when putting up the PR. The only regression I found was with the `seed` view which is fixed in this PR.

Please double test 🙏 .

src/component/list.js
erkarl: @valentinewallace updated - all lists now have sticky headers.

package.json
tanx: I tested the new dependency version. Looks like it breaks the seed view in storybook.

erkarl: This the output I'm getting in `erkarl:sticky-list-headers` branch:
asdf

Can you elaborate on what's broken here?

tanx: This is on my machine after installing the versions in the package.json

bildschirmfoto 2018-10-04 um 15 37 43

tanx: Ok I upgrade react/react-native-web in a separate PR and merged to master. @erkarl would you kindly rebase this PR on top of the master? thanks

erkarl:
@tanx rebased. Should we also squash the commits?

src/component/list.js
```diff
@@ -50,6 +50,7 @@ export class List extends Component {
         ref={component => (this.list = component)}
         dataSource={this.dataSource}
         renderHeader={this.props.renderHeader}
+        stickyHeaderIndices={[0]}
```
tanx: Since we're setting this globally for all `List` instances, we'll also need to add `style={{ backgroundColor: color.white }}` to the `ListHeader` in list-story in storybook.

erkarl: Done.

PR #734, opened 2018-10-16
Media queries

### LINKED ISSUE #576: Implement Loader - First Time mobile screen
https://invis.io/24NACMCBFM6

tanx:
Closes #576

### DIFF
--- a/src/view/loader.js
+++ b/src/view/loader.js
@@ -4,7 +4,7 @@ import { createStyles, maxWidth } from '../component/media-query';
 import Background from '../component/background';
 import { CopyOnboardText } from '../component/text';
 import MainContent from '../component/main-content';
-import { color } from '../component/style';
+import { color, invisible } from '../component/style';
 
 const baseStyles = {
   content: {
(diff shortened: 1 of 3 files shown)


src/view/loader.js
```diff
@@ -28,7 +29,18 @@ const styles = StyleSheet.create({
...
-});
+};
+
+const styles = createStyles(
+  baseStyles,
+
+  maxWidth(500, {
+    copy2: {
+      height: 0,
+      opacity: 0,
```
valentinewallace: I was able to get rid of this styling line (the opacity one) 
```suggestion
```

and it it still seems to look fine.

valentinewallace: I was able to get rid of this styling line (the opacity one) 
```suggestion
```

and it it still seems to look fine.

tanx:
> Looks good to me!
> 
> Some notes, I don't fully understand the code and it hasn't been updated in 2 years. On the other hand, it _looks_ simple and does seem to "just work."

Yeah, its simple but I had to massage the code a bit to work. So I just decided to import it as our own component.

> Approving (with one style simplification suggestion)! This is a great step for letting us stick to one codebase without sacrificing the user experience :D

Yay. Removing the opacity work find on iOS but it will still render them in chromium. I created a reusable `invisible` style that covers all cases and works on all platforms.

PR #735, opened 2018-10-17
Reset password

### LINKED ISSUE #498: Implement change password screens
https://invis.io/5XND89UHJKV
https://invis.io/CNND8AXFTB3
https://invis.io/M9ND8BB84PT
https://invis.io/28ND8BLYAVR
https://invis.io/2BND8BSYTD4
https://invis.io/5KND8E24CXR
https://invis.io/89ND8EDFUH4
https://invis.io/NCND8EVX6YG

valentinewallace:
Closes #498. Pretty big PR, so let me know if there's any larger architectural changes that should happen for first pass.


public/electron.js
```diff
@@ -129,6 +129,24 @@ function createWindow() {
...
+      lndPeerPort: LND_PEER_PORT,
+      logger: Logger,
+      lndArgs,
+    });
+  } catch (err) {
+    restartError = err;
+  }
+  event.sender.send('lnd-restart-error', { restartError });
+});
+
```
tanx: Not sure I understand why we need to restart lnd. Wouldn't it just be easier to restart the app?

src/action/index.js
```diff
@@ -34,7 +34,7 @@ export const log = new LogAction(store, ipc);
 export const nav = new NavAction(store);
 export const grpc = new GrpcAction(store, ipc);
 export const notify = new NotificationAction(store, nav);
-export const wallet = new WalletAction(store, grpc, db, nav, notify);
+export const wallet = new WalletAction(store, grpc, db, nav, notify, ipc);
```
tanx: We shouldn't inject electron's ipc into actions directly as this would make porting to mobile harder. Rather we should use the grpc actions to wrap ipc calls.

src/action/index.js
```diff
@@ -73,27 +73,32 @@ observe(store, 'firstStart', async () => {
...
  */
 observe(store, 'walletUnlocked', async () => {
-  await nap();
-  await grpc.closeUnlocker();
-  await grpc.initLnd();
+  if (store.walletUnlocked) {
+    await nap();
+    await grpc.closeUnlocker();
+    await grpc.initLnd();
+  }
```
tanx: This would be unnecessary if we just restarted the app.

src/action/index.js
```diff
@@ -73,27 +73,32 @@ observe(store, 'firstStart', async () => {
...
 observe(store, 'lndReady', () => {
-  wallet.getNewAddress();
-  wallet.pollBalances();
-  wallet.pollExchangeRate();
-  channel.update();
-  transaction.update();
-  transaction.subscribeTransactions();
-  transaction.subscribeInvoices();
-  info.pollInfo();
+  if (store.lndReady) {
```
tanx: Diff would be cleaner with `if (!store.lndReady) return;`.

src/action/wallet.js
```diff
@@ -261,6 +309,40 @@ class WalletAction {
...
+          restartError,
+        });
+      }
+      this._store.walletUnlocked = false;
+      this._store.lndReady = false;
+      await this._grpc.initUnlocker();
+      await this._grpc.sendUnlockerCommand('ChangePassword', {
+        current_password: toBuffer(currentPassword),
+        new_password: toBuffer(newPassword),
+      });
```
tanx: I see this is why we need to restart lnd 🤔

public/preload.js
```diff
@@ -6,7 +6,9 @@ const _ipcRenderer = require('electron').ipcRenderer;
 
 const filter = event => {
   if (
-    !/^(lnd)|(unlock)|(log)|(locale)|(open-url)[a-zA-Z_-]{0,20}$/.test(event)
+    !/^(lnd)|(unlock)|(restart-lnd-process)|(log)|(locale)|(open-url)[a-zA-Z_-]{0,20}$/.test(
+      event
+    )
```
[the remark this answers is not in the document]
valentinewallace: @tanx Meaning we'd accept all IPC events that are prefixed with `lnd-`?

src/action/wallet.js
valentinewallace: So the restart is acceptable?

src/action/index.js
valentinewallace: true, fixed!

public/preload.js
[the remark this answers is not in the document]
tanx: Yes. The plan is to prefix all ipc events with `lnd-`.

src/action/wallet.js
tanx: Looks like there's not way around it with the current lnd architecture @Roasbeef ?

public/electron.js
```diff
@@ -129,6 +129,24 @@ function createWindow() {
   });
 }
 
+ipcMain.on('restart-lnd-process', async event => {
```
tanx: Changin this to 'lnd-restart-process' would allow us to keep the regex filter the same. We should migrate other event names too, but no rush on that.

src/action/index.js
```diff
@@ -73,21 +73,25 @@ observe(store, 'firstStart', async () => {
...
  */
 observe(store, 'walletUnlocked', async () => {
-  await nap();
-  await grpc.closeUnlocker();
-  await grpc.initLnd();
+  if (store.walletUnlocked) {
+    await nap();
+    await grpc.closeUnlocker();
+    await grpc.initLnd();
+  }
```
tanx: Change to `if (!store. walletUnlocked) return;` like below.

public/electron.js
valentinewallace: sounds good, fixed

src/action/index.js
valentinewallace: fixed :)

PR #738, opened 2018-10-19
Update lnd commit

### LINKED ISSUE #668: Improve loader - syncing chain spinner

valentinewallace:
This updated commit should improve performance on startup for neutrino. Also should close #668. 

However, it breaks integration tests. The fix is to mine blocks before starting up `lnd`.

### DIFF
--- a/test/integration/action/action-integration.spec.js
+++ b/test/integration/action/action-integration.spec.js
@@ -14,6 +14,7 @@ import PaymentAction from '../../../src/action/payment';
 import InvoiceAction from '../../../src/action/invoice';
 import { nap, retry } from '../../../src/helper';
 import { EventEmitter } from 'events';
+import { BTCD_MINING_ADDRESS } from '../../../src/config';
 
 const {
   startLndProcess,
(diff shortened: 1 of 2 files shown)


test/integration/action/action-integration.spec.js
```diff
@@ -105,10 +106,12 @@ describe('Action Integration Tests', function() {
       isDev,
       logger,
       btcdSettingsDir: BTCD_SETTINGS_DIR,
+      miningAddress: BTCD_MINING_ADDRESS,
```
tanx: Why do we need to set the miningAddress? We were mining block before without it in the integration tests.

test/integration/action/action-integration.spec.js
```diff
@@ -105,10 +106,12 @@ describe('Action Integration Tests', function() {
...
     };
     btcdProcess = await startBtcdProcess(btcdArgs);
     await nap(NAP_TIME);
     await retry(() => isPortOpen(BTCD_PORT));
+    await mineBlocks({ blocks: 400, logger });
```
tanx: Curious why we need to mine block before even starting the lnd processes?

test/integration/action/action-integration.spec.js
valentinewallace: When? When I leave out that line, I get this error: 
deepinscreenshot_select-area_20181019125256

test/integration/action/action-integration.spec.js
valentinewallace: There's been a change to neutrino such that the lnd server won't start until the chain height is non-zero, because it considers zero to be "not caught up."

PR #746, opened 2018-10-23
Hide seen channels alert

### LINKED ISSUE #745: Remove red channel alert icon after user clicks to view channels page
deepinscreenshot_select-area_20181022173140

valentinewallace:
Closes #745. 

One of the pieces of feedback we received from multiple users over the course of the alpha launch was that the alert signifying a channel-related notification does not currently disappear after the user has seen it. This PR fixes that issue.


tanx:
> the alert signifying a channel-related notification does not currently disappear after the user has seen it

The question is do we want to remove it after the user has seen it or until there are channels in the channel list?

valentinewallace:
That seems a bit weird, seems like in that case we should be showing the notification when channels become present instead of while they're not present?

src/component/media-query.js
```diff
@@ -97,5 +97,5 @@ export const minAspectRatio = aspectRatioQuery(
   (target, aspectRatio) => target <= aspectRatio
 );
 export const aspectRatio = aspectRatioQuery(
-  (target, aspectRatio) => target == aspectRatio
+  (target, aspectRatio) => target === aspectRatio
```
tanx: Should already be fixed in the master.

test/unit/computed/channel.spec.js
```diff
@@ -49,6 +52,12 @@ describe('Computed Channels Unit Tests', () => {
       expect(store.showChannelAlert, 'to equal', true);
     });
 
+    it('should not show channel alert after navigating to view', () => {
+      nav.goChannels();
```
tanx: Isn't the flag set to false by the channel action `init()`?

valentinewallace: good point, fixed

valentinewallace:
I'm sort of unsure about this PR because once `channel.alertSeen` is set to true, it will never become false again*. Can close for now if you want. Could become more useful once we have a `channel updates` subscription in `lnd` + the app.

\* because the channels list only updates when the user navigates to the page, which will always cause `alertSeen` to be set to true.

tanx:
> * because the channels list only updates when the user navigates to the page, which will always cause `alertSeen` to be set to true.

Exactly. That was my thinking too. I feel like it's more of a feature than a bug that the alert is visible until auto pilot has opened a channel. Otherwise the user might think he can transact even though no channels exist. With the alert set, he'll at least know something is wrong and can open a channel manually.

> I'm sort of unsure about this PR because once `channel.alertSeen` is set to true, it will never become false again*. Can close for now if you want. Could become more useful once we have a `channel updates` subscription in `lnd` + the app.

Agreed. Let's close this for now :)

PR #751, opened 2018-10-24
Show closed channels in channels list.

### LINKED ISSUE #507: Display closed channels in the channel list using `ClosedChannels` grpc api

### LINKED ISSUE #549: Display closing transaction ID in [pending] closed channel detail

### LINKED ISSUE #610: Pending channels should sort to the top of the channel list when they become open
Currently, if pending channels at the bottom of the channel list become open, the list looks like:
```
[open channel]
[open channel]
[pending channel]
[pending channel]
[open channel]
[open channel]
```

Instead, channels that go from pending to open status should sort to the top when they become open.

valentinewallace:
Further, ensure they sort to the bottom.

Closes #507, closes #610, closes #549

### DIFF
--- a/src/action/channel.js
+++ b/src/action/channel.js
@@ -83,6 +83,7 @@ class ChannelAction {
       this.getPeers(),
       this.getChannels(),
       this.getPendingChannels(),
+      this.getClosedChannels(),
     ]);
   }
 
@@ -144,12 +145,12 @@ class ChannelAction {
       }));
       const pccs = response.pending_closing_channels.map(pcc => ({
         ...mapPendingAttributes(pcc.channel),
-        closingTxid: pcc.closing_txid,
+        closingTxId: pcc.closing_txid,
         status: 'pending-closing',
       }));
       const pfccs = response.pending_force_closing_channels.map(pfcc => ({
         ...mapPendingAttributes(pfcc.channel),
-        closingTxid: pfcc.closing_txid,
+        closingTxId: pfcc.closing_txid,
         limboBalance: pfcc.limbo_balance,
         maturityHeight: pfcc.maturity_height,
         blocksTilMaturity: pfcc.blocks_til_maturity,
@@ -163,6 +164,31 @@ class ChannelAction {
       this._store.pendingChannels = [].concat(pocs, pccs, pfccs, wccs);
     } catch (err) {
       log.error('Listing pending channels failed', err);
+    }
+  }
+
+  /**
+   * List the closed channels by calling the respective grpc api and updating
+   * the closed channels array in the global store.
+   * @return {Promise<undefined>}
+   */
+  async getClosedChannels() {
+    try {
+      const { channels } = await this._grpc.sendCommand('closedChannels');
+      this._store.closedChannels = channels.map(channel => ({
+        remotePubkey: channel.remote_pubkey,
+        capacity: parseSat(channel.capacity),
+        channelPoint: channel.channel_point,
+        fundingTxId: this._parseChannelPoint(channel.channel_point)
+          .funding_txid_str,
+        localBalance: parseSat(channel.settled_balance),
+        remoteBalance:
+          parseSat(channel.capacity) - parseSat(channel.settled_balance),
+        closingTxId: channel.closing_tx_hash,
+        status: 'closed',
+      }));
+    } catch (err) {
+      log.error('Listing closed channels failed', err);
     }
   }
 
--- a/src/computed/channel.js
+++ b/src/computed/channel.js
@@ -8,14 +8,15 @@ import { toAmountLabel, toCaps } from '../helper';
 const ComputedChannel = store => {
   extendObservable(store, {
     get computedChannels() {
-      const { channels, pendingChannels, settings } = store;
+      const { channels, pendingChannels, closedChannels, settings } = store;
       const c = channels ? channels.slice() : [];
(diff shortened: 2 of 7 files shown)


src/action/channel.js
```diff
@@ -100,7 +101,6 @@ class ChannelAction {
       const { channels } = await this._grpc.sendCommand('listChannels');
       this._store.channels = channels.map(channel => ({
         remotePubkey: channel.remote_pubkey,
-        id: channel.chan_id,
```
tanx: Should we remove this? Might be helpful to have the attribute internally even if we don't display it in the UI currently.

src/action/channel.js
```diff
@@ -166,6 +166,32 @@ class ChannelAction {
...
+        remotePubkey: channel.remote_pubkey,
+        capacity: parseSat(channel.capacity),
+        channelPoint: channel.channel_point,
+        fundingTxId: this._parseChannelPoint(channel.channel_point)
+          .funding_txid_str,
+        localBalance: parseSat(channel.settled_balance),
+        remoteBalance:
+          parseSat(channel.capacity) - parseSat(channel.settled_balance),
+        status: 'closed',
+      }));
```
tanx: Don't we want to display the `closingTxid` here as well?

src/computed/channel.js
```diff
@@ -8,14 +8,15 @@ import { toAmountLabel, toCaps } from '../helper';
...
       const p = pendingChannels ? pendingChannels.slice() : [];
-      const all = [].concat(c, p);
-      all.sort(
+      const cl = closedChannels ? closedChannels.slice() : [];
+      c.sort((a, b) => (a.active === b.active ? 0 : a.active ? -1 : 1));
+      p.sort(
         (a, b) => (a.status > b.status ? -1 : a.status < b.status ? 1 : 0)
       );
-      all.sort((a, b) => (a.active === b.active ? 0 : a.active ? -1 : 1));
+      let all = [].concat(c, p, cl);
```
tanx: Nice!

src/action/channel.js
valentinewallace: True, I changed it so the `details` page shows the closing tx ID for (pending) closed channels.

valentinewallace:
I updated it so that instead of showing a "close channel" button for channels that aren't cooperatively closeable, it shows a `DONE` button. I think this is a good intermediate until we enable force closes in the app (for offline or pending channels), lmk what you think.

src/view/channel-detail.js
```diff
@@ -49,9 +54,18 @@ const ChannelDetailView = ({ store, nav }) => (
...
+            style={styles.deleteBtn}
+            onPress={() => nav.goChannelDelete()}
+          >
+            <ButtonText style={styles.deleteBtnText}>CLOSE CHANNEL</ButtonText>
+          </Button>
+        ) : (
+          <Button style={styles.deleteBtn} onPress={() => nav.goChannels()}>
+            <ButtonText style={styles.deleteBtnText}>DONE</ButtonText>
+          </Button>
+        )}
```
tanx: Why do we need to display a `DONE` button when there is already an `X` at the top right of the modal?

Also, we should be aware that there are certain cases where the user may want to force close non-active channels. In any case this change doesn't belong in this PR IMHO.

tanx: To clarify, if you look at the current implementation of `closeSelectedChannel` in the channels action, it allows force closing. This change would remove that feature. If we want to change the UX for force closing, we should leave that for another PR.

valentinewallace: Ah that's my mistake, last time I checked force closing was never enabled. Fixed!

test/integration/action/action-integration.spec.js
```diff
@@ -458,24 +458,28 @@ describe('Action Integration Tests', function() {
...
     });
 
-    it('should list no channels after mining 6 blocks', async () => {
+    it('should list closed after mining 6 blocks', async () => {
       await mineAndSync({ blocks: 6 });
-      while (store1.pendingChannels.length) await nap(100);
-      expect(store1.computedChannels.length, 'to be', 0);
+      while (!store1.closedChannels.length) {
+        await nap(100);
+        channels1.update();
```
tanx: Shouldn't this be called inside the actions to make sure the store is updated?

test/integration/action/action-integration.spec.js
```diff
@@ -487,14 +491,19 @@ describe('Action Integration Tests', function() {
...
-    it('should list no channels after mining 6 blocks', async () => {
-      await mineAndSync({ blocks: 6 });
-      while (store1.pendingChannels.length) await nap(100);
-      expect(store1.computedChannels.length, 'to be', 0);
+    it('should list closed after mining more than 144 blocks', async () => {
+      await mineAndSync({ blocks: 144 });
+      while (store1.closedChannels.length < 2) {
+        await nap(100);
+        await mineAndSync({ blocks: 1 });
+        channels1.update();
```
tanx: Same here ...

test/integration/action/action-integration.spec.js
valentinewallace: Do you mean the polling?

tanx: You’re right. I tried call channel.update() on the channel close event and it didn’t list the closed channels yet. The channels are updated when navigating to the view anyway. Just not while the channel list is displayed. Not a big deal imho though.

PR #754, opened 2018-10-27
Upon password change cancel, navigate back to Settings instead of Home.

valentinewallace:
In the course of testing the app, I found that the `change password` flow redirects to Home instead of back to Settings if the user exits. This felt really unnatural/abrupt to me, so I decided to open a quick PR to fix it. 

Can close if you disagree, NBD.


tanx:
Closing in favor of #759

PR #772, opened 2018-11-03
Fix integration test flakes.

### LINKED ISSUE #761: Investigate why integration tests are flakey and make them more robust

valentinewallace:
Closes #761. 

There is [one odd flake](https://travis-ci.org/lightninglabs/lightning-app/jobs/449677596) that has a slightly different failure, but the other 17 or so had the same failure. 

I've been experiencing this flake locally for a while now, and I'm not sure what about `reset-password` caused it to suddenly surface in production. Regardless, this change fixed it for me locally so I think it should fix it on travis as well.

### DIFF
--- a/test/integration/action/test-util.js
+++ b/test/integration/action/test-util.js
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import net from 'net';
+import { nap } from '../../../src/helper';
 
 export const rmdir = path => {
   if (fs.existsSync(path)) {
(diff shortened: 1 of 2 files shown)


test/integration/action/test-util.js
```diff
@@ -24,3 +25,15 @@ export const isPortOpen = async port => {
...
+
+export const killProcess = async pid => {
+  let terminated = false;
+  while (!terminated) {
+    try {
+      process.kill(pid);
+      await nap(500);
+    } catch (e) {
+      terminated = true;
+    }
```
tanx: What if `process.kill()` is successful? Will it still throw an exception?

valentinewallace: If `process.kill()` is successful, then it will error on the next attempt because the process doesn't exist anymore. I couldn't find any other way to determine whether the process was actually dead. 

Tried `process.terminated`, tried `process.on(SIGINT)`/other event handlers. Those could probably work if I put in a `sleep` after they occur and then restarted `btcd`, but I wanted to avoid sleeps.

tanx: Ah ok. Let's see if this helps.

PR #781, opened 2018-11-08
Automatically focus first input during restore seed pagination.

### LINKED ISSUE #729: Autofocus is broken starting on the 2nd page of seed restore

valentinewallace:
Closes #729.

### DIFF
--- a/src/view/restore-seed.js
+++ b/src/view/restore-seed.js
@@ -52,7 +52,7 @@ const RestoreSeedView = ({ store, wallet }) => (
               wallet.setRestoreSeed({ word, index: seedIndex - 1 })
             }
             key={i}
-            autoFocus={i === 0}
+            autoFocus={i === 0 && store.wallet.autofocusRestoreInput}
             onSubmitEditing={() => wallet.initNextRestorePage()}
           />
         ))}
--- a/src/action/wallet.js
+++ b/src/action/wallet.js
@@ -53,6 +53,7 @@ class WalletAction {
    * @param {number} options.index The seed index
    */
   setRestoreSeed({ word, index }) {
+    this._store.wallet.autofocusRestoreInput = false;
     this._store.wallet.restoreSeed[index] = word;
   }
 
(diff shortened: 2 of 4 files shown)


src/component/input.js
```diff
@@ -20,19 +20,32 @@ const baseStyles = StyleSheet.create({
...
-    underlineColorAndroid="rgba(0,0,0,0)"
-    placeholderTextColor={color.greyPlaceholder}
-    {...props}
-  />
-);
+export class TextInput extends Component {
+  componentDidUpdate() {
+    const { autoFocus, value } = this.props;
+    value.length === 0 && autoFocus && this._input.focus();
+  }
```
tanx: Cleaned this up a bit in https://github.com/lightninglabs/lightning-app/pull/781/commits/f8010e922c04e7e7adbe40be6b72a82286e7ca02

valentinewallace:
I tried flipping `autoFocus` from true to false in `view/seed-verify`, but that still didn't refocus the correct input for me. This solution seems work across screens that use autofocus, but let me know if it doesn't work for you.

It's a bit hacky still :( maybe we should revisit this in a bit?

tanx:
@valentinewallace is this PR still relevant with the recent patch to textinput focus?

valentinewallace:
@tanx Yep, just tested and recovery input autofocus is still a bug.

src/view/restore-seed.js
```diff
@@ -52,7 +52,7 @@ const RestoreSeedView = ({ store, wallet }) => (
               wallet.setRestoreSeed({ word, index: seedIndex - 1 })
             }
             key={i}
-            autoFocus={i === 0}
+            autoFocus={i === 0 && store.wallet.autofocusRestoreInput}
```
tanx: @valentinewallace Not really sure I understand ... why does `store.wallet.autofocusRestoreInput` need to be false for the `setRestoreSeed()` case?

valentinewallace: Disclaimer: I don't fully know why this solution works.

It needs `autofocusRestoreInput` to flip from false to true when the user turns the page, or else the bug persists. I decided to set it to false in `setRestoreSeed()` but there may be a better place.

valentinewallace:
Replaced by #1106.

PR #788, opened 2018-11-13
Fix touchable list item

### LINKED ISSUE #787: Media query breaks desktop screens in electron production build
The settings list items don't render properly in the electron production build when using media queries :(

Looks like we'll either have to create a workaround or just build separate screen for desktop and mobile instead of using media queries.

This is currently preventing us from shipping an update since we've already merged #782 into the master.

tanx:
Our previous patches to the list-item for `view/select-seed.js` caused a rendering issue in the electron production build.

Closes #787


valentinewallace:
For me, the current `master` settings screen renders fine as far as I can tell (the separating lines look a bit finer). But, it also renders fine with these changes. So, merging! :)

PR #809, opened 2018-11-26
Grpc long number

### LINKED ISSUE #808: Use JS Number for long integers in grpc messages

tanx:
This improves compatibility between mobile and desktop grpc message parsing since mobile converts long integers to JS Numbers by default.

Closes #808

### DIFF
--- a/public/grpc-client.js
+++ b/public/grpc-client.js
@@ -63,7 +63,7 @@ module.exports.init = async function({
     protoPath = path.join(__dirname, '..', 'assets', 'rpc.proto');
     const options = {
       keepCase: false,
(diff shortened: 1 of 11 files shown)


tanx:
> Was there a reason we weren't always doing this?

Yeah the old proto parser converted long to strings by default and wasn't as configurable as the new one. Turns out this was unnecessary anyway since there are fewer total satoshis than `Number.MAX_SAFE_INTEGER` in JavaScript.  (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)

PR #826, opened 2018-12-05
assets/script: update install_lnd.sh to use go modules

Roasbeef:
In this commit, we update `install_lnd.sh` to use go modules. Along the
way, we also update the latest `lnd` and `btcd` commits to the latest
versions that can only be built with go modules.

### DIFF
--- a/assets/script/install_lnd.sh
+++ b/assets/script/install_lnd.sh
@@ -2,8 +2,8 @@
 
 # versions
 GO_TAG=1.11.1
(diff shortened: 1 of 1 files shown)


tanx:
@Roasbeef I also added `GO111MODULE=on` to the deployment script that builds the app https://github.com/lightninglabs/lightning-app/commit/fdf451416dfd513150d7d4abc411a6f4b3243f79

PR #835, opened 2018-12-14
Show inactive chans

### LINKED ISSUE #505: Display (in)active channel state in channel list

valentinewallace:
Closes #505.

### DIFF
--- a/src/computed/channel.js
+++ b/src/computed/channel.js
@@ -28,7 +28,14 @@ const ComputedChannel = store => {
     get channelBalanceOpenLabel() {
       const { channels, settings } = store;
       const sum = (channels || [])
-        .map(c => Number(c.localBalance))
+        .map(c => (c.active ? Number(c.localBalance) : 0))
+        .reduce((a, b) => a + b, 0);
+      return toAmountLabel(sum, settings);
+    },
+    get channelBalanceInactiveLabel() {
+      const { channels, settings } = store;
+      const sum = (channels || [])
+        .map(c => (!c.active ? Number(c.localBalance) : 0))
         .reduce((a, b) => a + b, 0);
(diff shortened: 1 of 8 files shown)


src/view/channel.js
```diff
@@ -171,6 +174,13 @@ const ChannelSummary = ({
         {channelBalanceOpenLabel} {unitLabel}
       </Text>
     </View>
+    <View style={summaryStyles.box}>
+      <Alert type="inactive" style={summaryStyles.alert} />
+      <Text style={summaryStyles.txt}>Inactive</Text>
+      <Text style={[summaryStyles.txt, summaryStyles.total]}>
+        {channelBalanceInactiveLabel} {unitLabel}
+      </Text>
+    </View>
```
tanx: I wouldn't add another summary here. Inactive balance should be included in the open balance.

valentinewallace: fixed :)

tanx: Really sorry. I confused this. Just retested and it's better with the separate summary. I reverted the commit.

src/computed/channel.js
```diff
@@ -35,6 +38,15 @@ const ComputedChannel = store => {
...
       const sum = (channels || [])
+        .filter(c => c.active)
+        .map(c => Number(c.localBalance))
+        .reduce((a, b) => a + b, 0);
+      return toAmountLabel(sum, settings);
+    },
+    get channelBalanceInactiveLabel() {
+      const { channels, settings } = store;
+      const sum = (channels || [])
+        .filter(c => !c.active)
```
tanx: @valentinewallace I also added `filter(c => !c.active)` here instead of the ternary expression. Seems a bit cleaner.

PR #839, opened 2018-12-17
Settings mobile

### LINKED ISSUE #601: Implement Settings mobile screen
https://invis.io/EDNS9WDZ3XR

### LINKED ISSUE #602: Implement Setting Units mobile screen
https://invis.io/RSNS9XDZN9K

tanx:
Closes #601, closes #602


src/component/list.js
```diff
@@ -117,6 +124,76 @@ ListHeader.propTypes = {
...
+SettingContent.propTypes = {
+  children: PropTypes.node,
+  style: View.propTypes.style,
+};
+
+//
+// Setting List
+//
+
+const baseLStyles = {
```
valentinewallace: I don't *really* understand why this component is necessary, but it doesn't work with out it so ¯\_(ツ)_/¯

src/component/list.js
```diff
@@ -117,6 +124,76 @@ ListHeader.propTypes = {
   style: View.propTypes.style,
 };
 
+//
+// Setting Content
```
valentinewallace: Maybe we should move all the `Setting` components into their own component file?

valentinewallace: Maybe we should move all the `Setting` components into their own component file?

tanx: > Maybe we should move all the `Setting` components into their own component file?

Good call. I moved the settings components into their own file in https://github.com/lightninglabs/lightning-app/pull/839/commits/f8c8ca456fed1f3071c8b0c0fe506c82cddd364a.

> Wow, that got rid of a ton of code! Very nice. Minor suggestion but feel free to merge.

Cool 🎉 Merging :)

src/component/list.js
tanx: This one is just a wrapper of the setting list items so that they can float freely in the desktop version.

PR #843, opened 2018-12-18
Deposit mobile

### LINKED ISSUE #594: Implement Deposit mobile screen
https://invis.io/CANS9NGWBEJ

valentinewallace:
Closes #594.

### DIFF
--- a/src/component/button.js
+++ b/src/component/button.js
@@ -414,7 +414,7 @@ const copyStyles = StyleSheet.create({
     flex: 1,
     height: 30,
     borderBottomColor: color.white,
-    borderBottomWidth: StyleSheet.hairlineWidth,
+    borderBottomWidth: 1,
   },
   text: {
     fontFamily: 'OpenSans Light',
@@ -453,7 +453,6 @@ const smallCopyStyles = StyleSheet.create({
     justifyContent: 'center',
     alignItems: 'center',
     height: 60,
-    borderRadius: 0,
     backgroundColor: color.purple,
     paddingLeft: 15,
     paddingRight: 15,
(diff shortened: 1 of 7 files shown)


src/view/deposit-mobile.js
```diff
@@ -0,0 +1,100 @@
...
+  },
+  btnWrapper: {
+    alignItems: 'center',
+  },
+  copyBtn: {
+    borderRadius: 0,
+    paddingLeft: 15,
+    paddingRight: 15,
+    backgroundColor: color.glas,
+  },
```
tanx: We should create another `SmallCopyButton` component that we can reuse in the invoice screen.

src/component/button.js
```diff
@@ -418,6 +418,7 @@ const copyStyles = StyleSheet.create({
   },
   text: {
     fontFamily: 'OpenSans Light',
+    textAlign: 'center',
```
tanx: ... changing the CopyButton for desktop shouldn't be necessary.

src/component/header.js
```diff
@@ -103,7 +103,7 @@ export const Title = ({ title = '', style, children }) => (
 
 Title.propTypes = {
   title: PropTypes.string,
-  style: View.propTypes.style,
+  style: Text.propTypes.style,
```
tanx: Good catch 👍

src/component/button.js
valentinewallace: fixed!

src/view/deposit-mobile.js
valentinewallace: Sounds good! should i open an issue for that or did you want it in the same PR?

PR #851, opened 2018-12-19
Fix iphone xs bottom color

### LINKED ISSUE #846: Fix lower background color in transaction detail and logs view

tanx:
Closes #846 

@valentinewallace this fixes the bottom background color on the iPhone XS.

### DIFF
--- a/src/view/transaction-detail-mobile.js
+++ b/src/view/transaction-detail-mobile.js
@@ -2,7 +2,7 @@ import React from 'react';
 import { StyleSheet, View } from 'react-native';
 import { observer } from 'mobx-react';
 import PropTypes from 'prop-types';
-import Background from '../component/background';
+import { SplitBackground } from '../component/background';
 import MainContent from '../component/main-content';
 import { Header, Title } from '../component/header';
 import { Button, BackButton } from '../component/button';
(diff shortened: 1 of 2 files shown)


src/view/transaction-detail-mobile.js
```diff
@@ -40,12 +40,13 @@ const styles = StyleSheet.create({
...
 
 const TransactionDetailView = ({ store, nav }) => (
-  <Background
+  <SplitBackground
     image={
       store.selectedTransaction.type === 'bitcoin'
         ? 'orange-gradient-bg'
         : 'purple-gradient-bg'
     }
+    bottom={color.whiteBg}
```
valentinewallace: To me this line doesn't do anything... So the problem this PR solves is only on the iphone XS?

tanx: > To me this line doesn't do anything... To clarify, the problem this PR solves is only on the iphone XS?

Oh sorry. Forgot to link the issue #846. This fixes the orange bar below the SafeAreaView.

tanx:
> To me, this PR doesn't change anything. But, it also doesn't hurt anything. Am I supposed to be seeing a difference? If not, feel free to merge.

I couldn't post a screenshot here for some reason. Sent it on app-dev on keybase. Ok merging.

PR #880, opened 2019-01-14
Display wait screen during sendCoins grpc call to prevent double click

### LINKED ISSUE #861: Prevent sending on-chain payment twice by switching to wait screen on pay button press

tanx:
Closes #861

### DIFF
--- a/test/unit/action/payment.spec.js
+++ b/test/unit/action/payment.spec.js
@@ -278,6 +278,7 @@ describe('Action Payments Unit Tests', () => {
       store.payment.address = 'some-address';
       grpc.sendCommand.withArgs('sendCoins').resolves();
       await payment.payBitcoin();
+      expect(nav.goWait, 'was called once');
       expect(grpc.sendCommand, 'was called with', 'sendCoins', {
         addr: 'some-address',
         amount: 1000,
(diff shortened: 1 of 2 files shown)


test/unit/action/payment.spec.js
```diff
@@ -289,7 +290,18 @@ describe('Action Payments Unit Tests', () => {
     it('should display notification on error', async () => {
       grpc.sendCommand.withArgs('sendCoins').rejects();
       await payment.payBitcoin();
+      expect(nav.goWait, 'was called once');
+      expect(nav.goPayBitcoinConfirm, 'was called once');
+      expect(notification.display, 'was called once');
+    });
+
+    it('should display error notification on timeout', async () => {
+      grpc.sendCommand.callsFake(() => nap(50));
```
valentinewallace: TIL about `callsFake`

tanx:
> Just imagine the support requests we're dodging right now.

Yeah, I'm glad we caught this one now 😳

> In case someone actually double-clicks, think it'd be worth opening an issue to disallow two spend attempts in the same 1 or 2 second period, in addition to these changes?

This fix should prevent that. The UI will navigate to the wait spinner before the grpc call is sent to the backend. So it shouldn't be possible to double click anymore.

PR #885, opened 2019-01-16
Notification mobile

### LINKED ISSUE #859: Display notifications on iOS

tanx:
Closes #859

### DIFF
--- a/src/component/notification.js
+++ b/src/component/notification.js
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { Animated, StyleSheet, View } from 'react-native';
+import { Animated, StyleSheet, View, SafeAreaView } from 'react-native';
 import PropTypes from 'prop-types';
 import { SmallPillButton } from './button';
 import { Text, H4Text } from './text';
@@ -12,6 +12,9 @@ import { color, font } from './style';
 //
 
 const barStyles = StyleSheet.create({
+  safe: {
+    backgroundColor: color.blackDark,
+  },
   bar: {
     alignSelf: 'stretch',
     flexDirection: 'row',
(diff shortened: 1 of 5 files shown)


src/component/notification.js
```diff
@@ -37,19 +40,21 @@ const barStyles = StyleSheet.create({
 
 export const NotificationBar = ({ notification, display, style }) =>
   notification && display ? (
-    <View style={[barStyles.bar, style]}>
-      <View style={barStyles.msgWrapper}>
-        <Alert type={notification.type} style={barStyles.alert} />
-        <Text style={barStyles.text}>{notification.message}</Text>
+    <SafeAreaView style={barStyles.safe}>
```
valentinewallace: IMO it looks a bit weird/abrupt to have the rest of the view shift downward when a notification appears.

src/config.js
```diff
@@ -4,7 +4,7 @@
 
 module.exports.RETRY_DELAY = 1000;
 module.exports.LND_INIT_DELAY = 5000;
-module.exports.NOTIFICATION_DELAY = 5000;
+module.exports.NOTIFICATION_DELAY = 2000;
```
valentinewallace: Hmm, some of them it feels like there's not enough time to click the `handler` but others with shorter messages it feels like a good amount of time. I wish there were an easier way to have the notifications stay until the user dismisses it :thinking:

valentinewallace: Maybe ntfns with a handler could have a longer delay somehow?

tanx: Good point. I retested this and 5 seconds is actually feels like a good amount and of time and should give the user enough time to register and react to the handler. So I reverted this change.

tanx:
> I misunderstood this issue: thought we were going to use native notifications. I prefer native notifications because we don't run into the problem of the page shifting down abruptly, or the notification disappearing too quickly or too slowly (which I felt were issues here -- though that could be subjective).

I tested using native notifications using expo's `LocalNotification` api. Unfortunately you cannot trigger native notifications on iOS while the app is displayed. Only while the app is in the background. Since we don't  have background processes yet. All our notification are currently while the app is open.

> Do you think these are problems? The code looks good, so I'm amenable to merging and then seeing what the team thinks and/or what user studies reveal later on.

Sgtm 👍

PR #890, opened 2019-01-18
Dynamically resize Home's balance label.

### LINKED ISSUE #879: Resize balance label to fit screen width on mobile (home screen)
@valentinewallace on iOS we can use https://facebook.github.io/react-native/docs/text#adjustsfontsizetofit

Let's ignore Android for now.

valentinewallace:
Closes #879.

### DIFF
--- a/src/component/label.js
+++ b/src/component/label.js
@@ -48,8 +48,12 @@ BalanceLabelNumeral.propTypes = {
   style: RNText.propTypes.style,
 };
 
-export const BalanceLabelUnit = ({ children, style }) =>
-  children ? <Text style={[balanceStyles.unit, style]}>{children}</Text> : null;
+export const BalanceLabelUnit = ({ children, style, ...props }) =>
+  children ? (
+    <Text style={[balanceStyles.unit, style]} {...props}>
+      {children}
+    </Text>
+  ) : null;
 
 BalanceLabelUnit.propTypes = {
   children: PropTypes.string,
--- a/src/component/label.js
+++ b/src/component/label.js
@@ -15,12 +15,11 @@ const balanceStyles = StyleSheet.create({
   numeral: {
     fontFamily: 'WorkSans ExtraLight',
     fontSize: font.sizeXXXL,
-    lineHeight: font.lineHeightXXXL,
+    lineHeight: null,
   },
   unit: {
     fontFamily: 'WorkSans Regular',
     fontSize: font.sizeL,
-    lineHeight: 60,
     marginLeft: 10,
   },
 });
@@ -35,7 +34,13 @@ BalanceLabel.propTypes = {
 };
 
 export const BalanceLabelNumeral = ({ children, style }) => (
-  <Text style={[balanceStyles.numeral, style]}>{children}</Text>
+  <Text
+    style={[balanceStyles.numeral, style]}
+    adjustsFontSizeToFit={true}
+    numberOfLines={1}
+  >
+    {children}
+  </Text>
 );
 
 BalanceLabelNumeral.propTypes = {
--- a/src/helper.js
+++ b/src/helper.js
@@ -321,3 +321,19 @@ export const generateArc = (x, y, radius, startAngle, endAngle) => {
     'Z',
   ].join(' ');
 };
+
+/**
+ * Calculate the appropriate top padding for the btc unit that sits next to the main
+ * home screen balance. This is needed because the height of the balance adjusts
+ * dynamically to its width.
+ * @param  {number} height The height of the balance.
+ * @return {number}        The amount of padding to put at the top of the unit.
+ */
+export const calculateTopPadding = height => {
+  if (height >= 80) {
+    return 15;
+  } else if (height >= 60) {
+    return 10;
+  }
+  return 5;
+};
--- a/src/helper.js
+++ b/src/helper.js
@@ -321,19 +321,3 @@ export const generateArc = (x, y, radius, startAngle, endAngle) => {
     'Z',
   ].join(' ');
 };
-
-/**
- * Calculate the appropriate top padding for the btc unit that sits next to the main
- * home screen balance. This is needed because the height of the balance adjusts
- * dynamically to its width.
- * @param  {number} height The height of the balance.
- * @return {number}        The amount of padding to put at the top of the unit.
- */
-export const calculateTopPadding = height => {
-  if (height >= 80) {
-    return 15;
-  } else if (height >= 60) {
-    return 10;
-  }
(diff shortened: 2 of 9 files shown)


src/view/home.js
```diff
@@ -92,7 +93,23 @@ const balanceStyles = StyleSheet.create({
...
+      width: 350,
+    },
+  }),
+
+  maxWidth(smallBreakWidth, {
+    wrapper: {
+      width: 250,
+    },
+  })
+);
```
tanx: Why does the wrapper need a width?

src/component/label.js
```diff
@@ -15,13 +15,15 @@ const balanceStyles = StyleSheet.create({
...
+    marginLeft: 10,
+    marginRight: 10,
   },
   unit: {
     fontFamily: 'WorkSans Regular',
     fontSize: font.sizeL,
-    lineHeight: 60,
-    marginLeft: 10,
+    paddingTop: 5,
+    marginRight: 10,
```
tanx: This break the alignment of the two labels at the top:

image

image

valentinewallace: Ahh yeah it's so tricky. When I make them start out at the same height, the unit shifts down awkwardly when the amount gets smaller in response to screen width. Working on this..

valentinewallace: @tanx Check it out now. A remaining issue is that when you have a balance with a lot of digits, the unit stops aligning at the top because its `marginTop` is hard coded to 15. 

The alternative I can think of would be a potentially complex calculation using DOM manipulation/`measure`: https://facebook.github.io/react-native/docs/direct-manipulation#measurecallback to dynamically set the `marginTop` instead. Maybe that's better? It seems like a bit of an edge case so I thought this might be acceptable.

src/view/home.js
tanx: Just tested removing the diff on home.js. It didn't change anything.

src/component/label.js
```diff
@@ -15,13 +15,15 @@ const balanceStyles = StyleSheet.create({
   numeral: {
     fontFamily: 'WorkSans ExtraLight',
     fontSize: font.sizeXXXL,
-    lineHeight: font.lineHeightXXXL,
+    lineHeight: null,
+    marginLeft: 10,
+    marginRight: 10,
```
tanx: The home screen already has 20 padding on the left and right. Margin shouldn't be required.

src/component/label.js
```diff
@@ -15,13 +15,15 @@ const balanceStyles = StyleSheet.create({
...
   },
   unit: {
     fontFamily: 'WorkSans Regular',
     fontSize: font.sizeL,
-    lineHeight: 60,
-    marginLeft: 10,
```
tanx: This `marginLeft: 10` is still required if `marginRight` on the numeral is removed.

src/component/label.js
tanx: > The alternative I can think of would be a potentially complex calculation using DOM manipulation/`measure`: https://facebook.github.io/react-native/docs/direct-manipulation#measurecallback to dynamically set the `marginTop` instead.

Could be worth a try.

Also worth noting the current solution works for fiat since the `$` symbol is already part of the string:
image

Just doesn't work for BTC:
image

valentinewallace: @tanx Are you sure the above UI bug isn't because you removed the `width` i set here, that you suggested I remove? https://github.com/lightninglabs/lightning-app/pull/890#discussion-diff-251703865R109

Because for me it scales when the width is set:
51284433_572653783208775_5728527671739023360_n

valentinewallace: It still clearly has problems but at least the font size of the number scales :sweat_smile:

valentinewallace: > Could be worth a try.

Okay, I'll try that solution today.

valentinewallace: @tanx Actually, this would be easy to solve without messing with DOM nodes if we had a monospace font (because we could just calculate `marginTop` for the unit based on the screen width and the number of characters in the balance).

Permission to ask Lee again?

src/component/label.js
valentinewallace: fixed :)

src/component/label.js
valentinewallace: fixed, thx

src/view/home.js
valentinewallace: I get this UI bug when I don't set the width in the wrapper:

51401843_294491867927029_6915343391798591488_n

valentinewallace: I think it's because `adjustFontSizeToFit` needs a reference point to know what font size to use?

tanx: @valentinewallace I think the issue with the btc unit behind the balance is causing `adjustFontSizeToFit` to fail. If you display then number as fiat it scales as intended.

src/view/home.js
```diff
@@ -92,26 +94,61 @@ const balanceStyles = StyleSheet.create({
...
+      <View style={balanceStyles.wrapper}>
+        <Button onPress={toggleDisplayFiat}>
+          <BalanceLabel>
+            <BalanceLabelNumeral>{channelBalanceLabel}</BalanceLabelNumeral>
+            <BalanceLabelUnit
+              onLayout={event => {
+                this.setState({ height: event.nativeEvent.layout.height });
+              }}
+              style={{ paddingTop: calculateTopPadding(this.state.height) }}
+            >
```
tanx: Views should not contain any logic. Is there a reason we need to `setState` here?

src/helper.js
```diff
@@ -321,3 +321,19 @@ export const generateArc = (x, y, radius, startAngle, endAngle) => {
...
+ * @return {number}        The amount of padding to put at the top of the unit.
+ */
+export const calculateTopPadding = height => {
+  if (height >= 80) {
+    return 15;
+  } else if (height >= 60) {
+    return 10;
+  }
+  return 5;
+};
```
tanx: Why is a component specific function in the generic helper module? We should probably move this to the corresponding component.

src/view/home.js
valentinewallace: Ah okay, so are you saying it's ok to leave the `width` as is?

src/view/home.js
valentinewallace: We need the height of the balance to dynamically calculate the `paddingTop` of the `btc` unit.

valentinewallace: fixed!

src/helper.js
valentinewallace: let me know if you like my solution!

tanx:
Closing in favor of #1295

PR #907, opened 2019-01-26
Fix bug such that if the app crashes while a user is restoring a wall…

### LINKED ISSUE #757: Handle app crash/quit before completion of recovery process
Currently, if a recovery is started but the app quits before it completes, then the app is unlocked, it won't properly go back into recovery mode. 

So there should be some persistence of the fact that the recovery did not complete, such that the next time we start the app we can set the `recovery_window` option during wallet unlock (so the backend knows to continue the recovery rescan).

valentinewallace:
…et, restoration will resume upon app restart.

Closes #757.

### DIFF
--- a/src/action/app-storage.js
+++ b/src/action/app-storage.js
@@ -26,9 +26,6 @@ class AppStorage {
         if (typeof this._store.settings[key] !== 'undefined') {
           this._store.settings[key] = state[key];
         }
-        if (key === 'restoring') {
-          this._store.settings[key] = JSON.parse(state[key]);
-        }
       });
     } catch (err) {
       log.error('Store load error', err);
--- a/src/action/app-storage.js
+++ b/src/action/app-storage.js
@@ -26,6 +26,9 @@ class AppStorage {
         if (typeof this._store.settings[key] !== 'undefined') {
           this._store.settings[key] = state[key];
         }
+        if (key === 'restoring') {
+          this._store.settings[key] = JSON.parse(state[key]);
+        }
       });
     } catch (err) {
       log.error('Store load error', err);
--- a/src/action/info.js
+++ b/src/action/info.js
@@ -38,7 +38,7 @@ class InfoAction {
         });
         this._store.percentSynced = this.calcPercentSynced(response);
       } else {
-        this._store.wallet.restoring = false;
+        this._store.settings.restoring = false;
         this._notification.display({
           type: 'success',
           msg: 'Syncing complete',
--- a/src/action/info.js
+++ b/src/action/info.js
@@ -38,6 +38,7 @@ class InfoAction {
         });
         this._store.percentSynced = this.calcPercentSynced(response);
       } else {
+        this._store.wallet.restoring = false;
         this._notification.display({
           type: 'success',
           msg: 'Syncing complete',
--- a/src/action/setting.js
+++ b/src/action/setting.js
@@ -43,6 +43,14 @@ class SettingAction {
   }
 
   /**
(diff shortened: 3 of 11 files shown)


tanx:
@valentinewallace rather than introducing complexity by adding addition state to work around the issue. I'd suggest we solve the underlaying issue.

src/action/app-storage.js
```diff
@@ -44,6 +48,10 @@ class AppStorage {
     try {
       const state = JSON.stringify(this._store.settings);
       await this._AsyncStorage.setItem('settings', state);
+      await this._AsyncStorage.setItem(
+        'restoringWallet',
+        this._store.wallet.restoring
+      );
```
tanx: We currently only persist the settings object and should keep it that way for consistency's sake. If additional attributes are added, they should be added on the settings object.

src/action/info.js
```diff
@@ -38,6 +38,7 @@ class InfoAction {
         });
         this._store.percentSynced = this.calcPercentSynced(response);
       } else {
+        this._store.wallet.restoring = false;
```
tanx: When is `store.wallet.restoring` set to true?

src/action/wallet.js
```diff
@@ -403,6 +403,7 @@ class WalletAction {
       this._nav.goWait();
       await this._grpc.sendUnlockerCommand('UnlockWallet', {
         walletPassword: toBuffer(walletPassword),
+        recoveryWindow: this._store.wallet.restoring ? 250 : 0,
```
tanx: Perhaps I'm missing something, but won't `recoveryWindow ` always be `0`since `store.wallet.restoring` is never set to true?

src/action/app-storage.js
valentinewallace: Changed `restoring` to live in the `settings` part of the `store`.

src/action/info.js
valentinewallace: On the `select-seed` view, when the user selects that they're restoring a wallet.

src/action/app-storage.js
```diff
@@ -26,6 +26,9 @@ class AppStorage {
         if (typeof this._store.settings[key] !== 'undefined') {
           this._store.settings[key] = state[key];
         }
+        if (key === 'restoring') {
+          this._store.settings[key] = JSON.parse(state[key]);
+        }
```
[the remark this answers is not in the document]
valentinewallace: Otherwise the property is parsed as a string, so this "casts" it to a boolean.

src/action/info.js
valentinewallace: Yes, sorry thought i fixed that

src/action/setting.js
```diff
@@ -56,6 +56,14 @@ class SettingAction {
...
     }
   }
+
+  /**
+   * Set whether or not we're restoring the wallet.
+   * @param {boolean} options.restoring Whether or not we're restoring.
+   */
+  setRestoringWallet({ restoring }) {
+    this._store.settings.restoring = restoring;
+  }
```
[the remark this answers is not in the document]
valentinewallace: fixed

src/action/app-storage.js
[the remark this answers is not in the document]
valentinewallace: Is there a better way? :grimacing:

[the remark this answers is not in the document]
tanx: > Otherwise the property is parsed as a string, so this "casts" it to a boolean.

Really? Why was that not an issue with `store.settings.displayFiat`? I mean the JSON is already parse in line 24? Or am I missing something?

PR #909, opened 2019-01-29
Enable users to connect to a remote lnd node.

### LINKED ISSUE #823: expose command line option for app to connect to remote lnd node
IIUC, there are two ways to use the application atm:
  1. Fully integrate lnd+neutrino mode. This spins up lnd locally, passing the arguments to use the neutrino mode, and also connect to the initial bootstrap cluster (on testnet). We primarily do releases to provide users with updated versions for this mode. Atm, this mode is _testnet only_, as there're some backend items within `lnd` and `neutrino` that need to be updated. 
  2. Using the command line, connect to a remote `btcd`, with and `lnd` instance spun up locally. IIRC, this was added earlier on before testnet neutrino's sync was stable enough. To my knowledge, not many users use this mode currently, as a testnet full node is pretty light, and option #1 now exists. 

I propose we add a 3rd operating mode: application running locally that connects to a _remote_ `lnd` node. This provide a simple interface for all existing routing node operators (casual and hardcore) that use `lnd` today. The current testnet deployment while functional is a bit limiting, as `testnet` is simply less "fun" than mainnet. Additionally, many services (regrettably) seem to now be skipping testnet all together and jumping directly to mainnet. By allowing users to connect to their _existing_ routing nodes, we're able to get some feedback from a slightly more advanced group (some set up to get this working compared to a double click) and also start to grow the user base of the application ahead of the full desktop and then later neutrino releases. 

The command+arguments on the cli would look something like: 
```
./Lightning.app --remote_connect=true --bitcoin.mainnet --rpcserver=roasbeef.computer:9099 --adminmacaroon=localpath.macaroon --tlscert=localpath.cert
```
The items we need are:
  * mainnet vs testnet
  * where to hit the rpc server at (will no longer be local host)
  * the admin macaroon (will no longer read from disk, will be specified in hex form)
  * the cert so we can connect to the remote lnd

In order for this to work for the users, they may need to generate a new `rpc.cert` that allows connections from sources other than `localhost`. They can do this with the `--tlsextraip` argument. They'll also need to modify `--rpclisten` to be `0.0.0.0` instead of `localhost`

With the above implemented, rather than spinning up a local lnd instance locally, the app instead attempts to connect out to the remote `lnd` instance (if those args are set on the command line). This would be the only functional change in this mode other than the lack of logs (as we don't get the logs over RPC).

valentinewallace:
Closes #823.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,15 @@ To run the packaged version of the app e.g. for macOS run:
 ./dist/mac/Lightning.app/Contents/MacOS/Lightning --bitcoin.node=bitcoind --bitcoind.rpcuser=kek --bitcoind.rpcpass=kek --bitcoind.zmqpubrawblock=tcp://127.0.0.1:28332 --bitcoind.zmqpubrawtx=tcp://127.0.0.1:28333
 ```
 
+### Connect to a remote lnd node
+1. In your remote node's `lnd.conf`, set `rpclisten=0.0.0.0:10006` and `tlsextraip=<insert your node's IP address here>`, then restart the node.
+
+2. Copy the `tls.cert` and `tls.key` from your remote node's `.lnd` into the appropriate folder for your platform (specified in the section below).
+
+3. Copy `admin.macaroon` from the remote node into `lnd/data/chain/bitcoin/<network>` (in the same folder you put the `tls` files).
+
+4. When starting the app, add the flag `--lndip=<your remote node's IP address>`. Note that the node must be locked when the app connects to it.
+
 ### Lnd data and logs
 Lnd data and logs are written to the following locations in production:
 
--- a/public/electron.js
+++ b/public/electron.js
@@ -6,6 +6,7 @@ const url = require('url');
 const isDev = require('electron-is-dev');
 const log = require('electron-log');
 const { startLndProcess, startBtcdProcess } = require('./lnd-child-process');
+const { parseCliArg } = require('../public/helper');
 const grcpClient = require('./grpc-client');
 const {
   PREFIX_NAME,
@@ -37,6 +38,7 @@ const btcdSettingsDir = path.join(isDev ? 'data' : userDataPath, 'btcd');
 const lndArgs = process.argv.filter(a =>
   /(^--bitcoin)|(^--btcd)|(^--neutrino)/.test(a)
 );
+let lndIP = parseCliArg('lndip');
 
 // Keep a global reference of the window object, if you don't, the window will
 // be closed automatically when the JavaScript object is garbage collected.
@@ -125,12 +127,21 @@ function createWindow() {
   grcpClient.init({
     ipcMain,
     lndSettingsDir,
+    lndIP,
     lndPort: LND_PORT,
     network: isDev ? 'simnet' : NETWORK,
   });
 }
 
 ipcMain.on('lnd-restart-process', async event => {
+  if (lndIP) {
+    event.sender.send(
+      'lnd-restart-error',
+      'Unable to reset password of remote lnd node.'
+    );
+    return;
+  }
+
   lndProcess && lndProcess.kill('SIGINT');
   let restartError;
   try {
@@ -199,7 +210,7 @@ app.on('ready', () => {
   initAutoUpdate();
   createWindow();
   initApplicationMenu();
-  startLnd();
+  if (!lndIP) startLnd();
 });
 
 // Quit when all windows are closed.
--- a/public/electron.js
+++ b/public/electron.js
@@ -6,7 +6,7 @@ const url = require('url');
 const isDev = require('electron-is-dev');
 const log = require('electron-log');
 const { startLndProcess, startBtcdProcess } = require('./lnd-child-process');
-const { parseCliArg } = require('../public/helper');
+const { parseCliArg } = require('./helper');
 const grcpClient = require('./grpc-client');
 const {
   PREFIX_NAME,
--- a/public/grpc-client.js
+++ b/public/grpc-client.js
@@ -48,6 +48,7 @@ function getMacaroonCreds(lndSettingsDir, network) {
 
 module.exports.init = async function({
   ipcMain,
+  lndIP = 'localhost',
   lndPort,
   lndSettingsDir,
   network,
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -104,3 +104,12 @@ Lnd data and logs are written to the following locations in production:
 * **Linux:** `~/.config/lightning-app/lnd`
 * **OSX:** `~/Library/Application Support/lightning-app/lnd`
 * **Windows:** `%USERPROFILE%\AppData\Roaming\lightning-app\lnd`
+
+### Connect to a remote lnd node
+1. In your remote node's `lnd.conf`, set `rpclisten=0.0.0.0:10006` and `tlsextraip=<insert your node's IP address here>`, then restart the node.
+
+2. Copy the `tls.cert` and `tls.key` from your remote node's `.lnd` into the appropriate folder for your platform (specified in the section above).
```
tanx: Will I overwrite my `tls.cert` and `tls.key`? Would that prevent the app from running in bundled mode?

public/electron.js
```diff
@@ -37,6 +37,13 @@ const btcdSettingsDir = path.join(isDev ? 'data' : userDataPath, 'btcd');
 const lndArgs = process.argv.filter(a =>
   /(^--bitcoin)|(^--btcd)|(^--neutrino)/.test(a)
 );
+let lndIP = 'localhost';
+process.argv.filter(a => {
+  if (/(^--lndip)/.test(a)) {
+    let split = a.split('=');
+    if (split.length > 1) lndIP = split[1];
+  }
+});
```
tanx: Move cli arg parsing to a function in `helpers.js`. This should be something like:
```
let lndIP = parseCliArg('lndip');
```

public/grpc-client.js
```diff
@@ -48,6 +48,7 @@ function getMacaroonCreds(lndSettingsDir, network) {
 
 module.exports.init = async function({
   ipcMain,
+  lndIP,
```
tanx: Set default value `lndIP = 'localhost'`

public/electron.js
```diff
@@ -199,7 +215,7 @@ app.on('ready', () => {
   initAutoUpdate();
   createWindow();
   initApplicationMenu();
-  startLnd();
+  if (lndIP === 'localhost') startLnd();
```
tanx: Change to `if (!lndIP) startLnd();`

public/electron.js
```diff
@@ -125,12 +132,21 @@ function createWindow() {
...
     ipcMain,
     lndSettingsDir,
+    lndIP,
     lndPort: LND_PORT,
     network: isDev ? 'simnet' : NETWORK,
   });
 }
 
 ipcMain.on('lnd-restart-process', async event => {
+  if (lndIP !== 'localhost') {
```
tanx: Change to `if (lndIP) {`

test/integration/action/action-integration.spec.js
```diff
@@ -140,12 +141,14 @@ describe('Action Integration Tests', function() {
...
     await grcpClient.init({
       ipcMain: ipcMainStub1,
+      lndIP,
       lndPort: LND_PORT_1,
       lndSettingsDir: LND_SETTINGS_DIR_1,
       network: NETWORK,
     });
     await grcpClient.init({
       ipcMain: ipcMainStub2,
+      lndIP,
```
tanx: Changes to this file shouldn't be necessary if the default value is set in `grpc-client.js`.

README.md
valentinewallace: Yeah it's a problem. I think I'm going to change it to accept CLI arguments for the path to the certs and the macaroons instead of telling the user to put them in a specific place. LMK if that works.

tanx: Also, we should add this section in the read above `### Lnd data and logs` since it goes together with the different modes of operation.

tanx: Cool! Ok then it shouldn't be an issue 👍

valentinewallace: ok sounds good, fixed the order of the README

public/electron.js
valentinewallace: fixed

public/grpc-client.js
valentinewallace: good point, fixed. could do the same with `lndPort`

public/electron.js
valentinewallace: fixed

public/electron.js
valentinewallace: fixed

public/electron.js
```diff
@@ -6,6 +6,7 @@ const url = require('url');
 const isDev = require('electron-is-dev');
 const log = require('electron-log');
 const { startLndProcess, startBtcdProcess } = require('./lnd-child-process');
+const { parseCliArg } = require('../public/helper');
```
tanx: Should be `require('./helper');`

valentinewallace: true. fixed

Roasbeef:
As implemented this will conflict with a local application on the same network+chain since it relies on overriding the existing on disk cert and macaroon. In the future, this will prevent users from using the local bundled application along with using it in detached node mode. There's also no UI level indicator w.r.t which network the application is running on. I foresee this causing user confusion down the line.

README.md
```diff
@@ -98,6 +98,15 @@ To run the packaged version of the app e.g. for macOS run:
 ./dist/mac/Lightning.app/Contents/MacOS/Lightning --bitcoin.node=bitcoind --bitcoind.rpcuser=kek --bitcoind.rpcpass=kek --bitcoind.zmqpubrawblock=tcp://127.0.0.1:28332 --bitcoind.zmqpubrawtx=tcp://127.0.0.1:28333
 ```
 
+### Connect to a remote lnd node
+1. In your remote node's `lnd.conf`, set `rpclisten=0.0.0.0:10006` and `tlsextraip=<insert your node's IP address here>`, then restart the node.
+
+2. Copy the `tls.cert` and `tls.key` from your remote node's `.lnd` into the appropriate folder for your platform (specified in the section below).
+
+3. Copy `admin.macaroon` from the remote node into `lnd/data/chain/bitcoin/<network>` (in the same folder you put the `tls` files).
```
Roasbeef: The location of this folder is platform specific. The notes (as implemented right now) should point the user to the directory on their platform.

Roasbeef:
Referring back to the desired launch command on the [original issue](https://github.com/lightninglabs/lightning-app/issues/823): 

```
./Lightning.app --remote_connect=true --bitcoin.mainnet --rpcserver=roasbeef.computer:9099 --adminmacaroon=localpath.macaroon --tlscert=localpath.cert
```

If it were implemented as illustrated above, the user wouldn't need to make new files on their file system. They would also be able to use the remove connect feature with both testnet and mainnet.

PR #935, opened 2019-02-06
Log mobile

### LINKED ISSUE #924: Implement action/log-mobile.js that uses new grpc-mobile logs event

tanx:
Closes #924

### DIFF
--- a/src/action/log.js
+++ b/src/action/log.js
@@ -9,6 +9,7 @@ import { MAX_LOG_LENGTH } from '../config';
 
 let _store;
 let _ipc;
+let _printErrObj;
 
 /**
  * Log an info event e.g. when something relevant but non-critical happens.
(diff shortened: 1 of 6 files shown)


test/unit/action/log.spec.js
```diff
@@ -97,10 +100,20 @@ describe('Action Logs Unit Tests', () => {
...
     });
+
+    describe('error()', () => {
+      it('should call console.error with all args', () => {
+        new LogAction(store, ipc, false);
+        const err = new Error('bar');
+        sandbox.stub(console, 'error');
+        log.error('foo', err);
+        expect(console.error, 'was called with', 'foo', err);
+        expect(store.logs.length, 'to equal', 56);
```
valentinewallace: Is it `56` because that's the length of the stringified `Error('bar')`?

src/action/log.js
```diff
@@ -36,7 +37,13 @@ export function error(...args) {
   pushLogs(''); // newline
   pushLogs(`ERROR: ${args[0]}`);
   for (let i = 1; i < args.length; i++) {
-    pushLogs(JSON.stringify(args[i], null, '    '));
+    pushLogs(
+      JSON.stringify(
```
valentinewallace: Is this reason for this that on mobile, errors are already objects? Why not always print the `stringify`'d object?

test/unit/action/log.spec.js
tanx: just a diff str length to be checked

src/action/log.js
tanx: Error objects have different formating on mobile and desktop. Since the grcp layer throws difference types. Just making sure we print the most readable version here.

valentinewallace: Ahh okay. LGTM then!

PR #961, opened 2019-02-11
Set high zIndex on button component

### LINKED ISSUE #960: Back button does not work on seed-mobile view
It seems the button is covered by something and does not have a high enough `z-index` in the UI.

tanx:
Closes #960


valentinewallace:
Works as advertised!

PR #974, opened 2019-02-13
Unified balance

### LINKED ISSUE #920: Display a single unified balance on the home screen
I've been thinking about our discussion regarding the home screen and the confusion and usability challenges we've seen around the balance numbers. None of the wording `Pending Deposit` or `Chain Deposit` has worked really well. My proposal is to display a single unified balance on the home screen that includes all on-chain as well as off-chain funds. To make sure users know if they can start doing lightning transactions, the balance should display a small indicator next to it that lights up when the app is ready to do lightning transactions.

My reasoning goes as follows:

1. In most cases it doesn't matter to the user how big the on-chain/off-chain balances are. Once there is a single open channel, the user can start spending.

2. Autopilot currently allocates 95% of funds to channels anyway. So the the unified number would be pretty close to the off-chain funds.

3. For the case that the user wants to transfer all funds out via a "wipe" on-chain transaction, we will provide them with a switch for autopilot in the settings.

What do you think @Roasbeef @bryanvu @valentinewallace ?

tanx:
Closes #920 

@valentinewallace Let's wait with merging until after we get feedback from testers if this is an improvement for them. But feel free to give a first once over review beforehand.


src/action/transaction.js
```diff
@@ -63,7 +63,7 @@ class TransactionAction {
         amount: transaction.amount,
         fee: transaction.totalFees,
         confirmations: transaction.numConfirmations,
-        status: transaction.numConfirmations < 1 ? 'unconfirmed' : 'confirmed',
+        status: transaction.numConfirmations < 3 ? 'unconfirmed' : 'confirmed',
```
valentinewallace: Note: in the past, we've gotten some feedback when people are able to send payments through channels that are displayed as `pending open`.

src/action/channel.js
```diff
@@ -87,6 +85,14 @@ class ChannelAction {
     ]);
   }
 
+  /**
+   * Poll the channels in the background since there is no streaming grpc api
+   * @return {Promise<undefined>}
+   */
+  async pollChannels() {
+    await poll(() => this.update());
```
valentinewallace: I think it would be a good idea to increase the polling time out (at least for channels) to 5 or 10 seconds. If this helps people's overheating problems, we know that the polling was the cause.

src/computed/channel.js
```diff
@@ -35,40 +35,59 @@ const ComputedChannel = store => {
       });
       return all;
     },
-    get channelBalanceOpenLabel() {
-      const { channels, settings } = store;
-      const sum = (channels || [])
+    get channelBalanceOpenSatoshis() {
```
valentinewallace: FWIW none of the unit tests cover this property or the other `.*Satoshi` computed channel properties. But, I don't think this is a huge problem.

src/computed/wallet.js
```diff
@@ -31,6 +35,15 @@ const ComputedWallet = store => {
       const { settings } = store;
       return !settings.displayFiat ? UNITS[settings.unit].display : null;
     },
+    get channelPercentageLabel() {
```
valentinewallace: :+1:

src/view/home.js
```diff
@@ -41,21 +41,26 @@ const HomeView = ({
...
+    totalBalanceLabel,
+    unitLabel,
+    channelStatus,
+    channelPercentageLabel,
+  } = store;
   return (
     <Background image="purple-gradient-bg">
       <HomeHeader
-        goChannels={() => channel.init()}
+        goDeposit={() => nav.goDeposit()}
```
valentinewallace: I got pretty confused when trying to find where the channels page went. I don't find clicking `95% on Lightning` very discoverable. 

One vote for the header going back to how it was before, or at least separating this feature into a different PR for further discussion :)

src/view/home.js
```diff
@@ -41,21 +41,26 @@ const HomeView = ({
...
       />
-      <QrCodeSeparator goDeposit={() => nav.goDeposit()} />
       <MainContent style={styles.content}>
         <BalanceDisplay
-          depositLabel={depositLabel}
-          channelBalanceLabel={channelBalanceLabel}
+          totalBalanceLabel={totalBalanceLabel}
           unitLabel={unitLabel}
+          channelStatus={channelStatus}
+          channelPercentageLabel={channelPercentageLabel}
```
valentinewallace: Totally love the way this looks! Way clearer than any of the other balance ideas we had.

src/view/home.js
```diff
@@ -87,36 +92,45 @@ const balanceStyles = StyleSheet.create({
...
       <BalanceLabel>
-        <BalanceLabelNumeral>{channelBalanceLabel}</BalanceLabelNumeral>
+        <BalanceLabelNumeral>{totalBalanceLabel}</BalanceLabelNumeral>
         <BalanceLabelUnit>{unitLabel}</BalanceLabelUnit>
       </BalanceLabel>
-      <H4Text style={balanceStyles.smallLabel}>Chain Deposit</H4Text>
-      <SmallBalanceLabel unit={unitLabel}>{depositLabel}</SmallBalanceLabel>
+    </Button>
+    <Button onPress={goChannels} style={balanceStyles.percentBtn}>
+      <Alert type={channelStatus} style={balanceStyles.alert} />
```
valentinewallace: I like how we're reusing the alerts. nice "design language"

test/unit/action/channel.spec.js
```diff
@@ -85,6 +81,15 @@ describe('Action Channels Unit Tests', () => {
     });
   });
 
+  describe('pollChannels()', () => {
```
valentinewallace: :+1:

src/action/transaction.js
tanx: Mmmh. That's wierd. Let's watch out for that. This change only affects the transaction list status.

src/view/home.js
tanx: > I don't find clicking `95% on Lightning` very discoverable.

I think we agreed to test this and get feedback from users as it's released. We can always iterate and improve on this later.

src/action/channel.js
tanx: Fixed this in b4c3e23

PR #1000, opened 2019-03-06
Toggle autopilot

### LINKED ISSUE #925: Toggle autopilot under settings
We should add a toggle button in the settings to disable autopilot so users can transfer their funds out.

valentinewallace:
Closes #925.

@tanx NOTE: we shouldn't merge this until the cluster is upgraded, since app users running this new `lnd` commit will reject the cluster as is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ To run the packaged version of the app e.g. for macOS run:
 
 The app is configured for testnet by default but you can opt-in to mainnet if you have a btcd node running. Be aware that this is currently still experimental:
 ```
-./dist/mac/Lightning.app/Contents/MacOS/Lightning --bitcoin.mainnet --bitcoin.node=neutrino --neutrino.connect=127.0.0.1 --autopilot.active
+./dist/mac/Lightning.app/Contents/MacOS/Lightning --bitcoin.mainnet --bitcoin.node=neutrino --neutrino.connect=127.0.0.1
 ```
 
 ### Starting the Packaged App (full node)
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ See the `ToDo (next release)` column on our [project board](https://github.com/l
 ```
 git clone https://github.com/lightningnetwork/lnd $GOPATH/src/github.com/lightningnetwork/lnd
 cd $GOPATH/src/github.com/lightningnetwork/lnd
-make && make install tags=experimental
+make && make install tags="experimental autopilotrpc"
 ```
 
 #### Install btcd
--- a/assets/autopilot.proto
+++ b/assets/autopilot.proto
@@ -1,0 +1,75 @@
+syntax = "proto3";
+
+package autopilotrpc;
+
+option go_package = "github.com/lightningnetwork/lnd/lnrpc/autopilotrpc";
+
+// Autopilot is a service that can be used to get information about the current
+// state of the daemon's autopilot agent, and also supply it with information
+// that can be used when deciding where to open channels.
+service Autopilot {
+    /**
+    Status returns whether the daemon's autopilot agent is active.
+    */
+    rpc Status(StatusRequest) returns (StatusResponse);
+
+    /**
+    ModifyStatus is used to modify the status of the autopilot agent, like
+    enabling or disabling it.
+    */
+    rpc ModifyStatus(ModifyStatusRequest) returns (ModifyStatusResponse);
+
+    /**
+    QueryScores queries all available autopilot heuristics, in addition to any
+    active combination of these heruristics, for the scores they would give to
+    the given nodes.
+    */
+    rpc QueryScores(QueryScoresRequest) returns (QueryScoresResponse);
+
+    /**
+    SetScores attempts to set the scores used by the running autopilot agent,
+    if the external scoring heuristic is enabled.
+    */
+    rpc SetScores(SetScoresRequest) returns (SetScoresResponse);
+}
+
+message StatusRequest{
+}
+
+message StatusResponse{
+    /// Indicates whether the autopilot is active or not.
+    bool active = 1 [json_name = "active"];
+}
+
+message ModifyStatusRequest{
+    /// Whether the autopilot agent should be enabled or not.
+    bool enable = 1 [json_name = "enable"];
+}
+
+message ModifyStatusResponse {}
+
+message QueryScoresRequest{
+    repeated string pubkeys = 1 [json_name = "pubkeys"];
+}
+
+message QueryScoresResponse {
+    message HeuristicResult {
+        string heuristic = 1 [json_name = "heuristic"];
+        map<string, double> scores= 2 [json_name = "scores"];
+    }
+
+    repeated HeuristicResult results = 1 [json_name = "results"];
+}
+
+message SetScoresRequest{
+    /// The name of the heuristic to provide scores to.
+    string heuristic = 1 [json_name = "heuristic"];
+
+    /**
+    A map from hex-encoded public keys to scores. Scores must be in the range
+    [0.0, 1.0].
+    */
+    map<string, double> scores = 2 [json_name = "scores"];
+}
+
+message SetScoresResponse {}
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "build-storybook": "build-storybook -s public",
     "build-icon": "svgr --native --icon --filename-case kebab -d src/asset/icon src/asset/icon",
     "build-img": "svgr --native --filename-case kebab -d src/asset/img src/asset/img",
-    "protobuf": "pbjs assets/rpc.proto -t static-module -w es6 -o assets/rpc.js && pbjs assets/autopilot.proto -t static-module -w es6 -o assets/autopilot.js"
+    "protobuf": "pbjs assets/rpc.proto -t static-module -w es6 -o assets/rpc.js"
   },
   "dependencies": {
     "@grpc/proto-loader": "^0.3.0",
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "files": [
       "build/**/*",
       "node_modules/**/*",
-      "assets/rpc.proto",
+      "assets/*.proto",
       "src/config.js"
     ],
     "directories": {
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "build-storybook": "build-storybook -s public",
     "build-icon": "svgr --native --icon --filename-case kebab -d src/asset/icon src/asset/icon",
     "build-img": "svgr --native --filename-case kebab -d src/asset/img src/asset/img",
-    "protobuf": "pbjs assets/rpc.proto -t static-module -w es6 -o assets/rpc.js"
+    "protobuf": "pbjs assets/rpc.proto -t static-module -w es6 -o assets/rpc.js && pbjs assets/autopilot.proto -t static-module -w es6 -o assets/autopilot.js"
   },
   "dependencies": {
     "@grpc/proto-loader": "^0.3.0",
--- a/public/grpc-client.js
+++ b/public/grpc-client.js
@@ -50,6 +50,12 @@ module.exports.init = async function({
   lndSettingsDir,
   network,
 }) {
+  let credentials;
+  let protoPath;
+  let lnrpc;
+  let unlocker;
+  let lnd;
+  let autopilot;
   const protoOptions = {
     keepCase: false,
     longs: Number,
@@ -57,13 +63,6 @@ module.exports.init = async function({
     defaults: true,
     oneofs: true,
   };
-  let credentials;
-  let protoPath;
-  let atplProtoPath;
-  let lnrpc;
-  let unlocker;
-  let lnd;
-  let autopilot;
 
   ipcMain.on('unlockInit', async event => {
     credentials = await getCredentials(lndSettingsDir);
@@ -94,8 +93,8 @@ module.exports.init = async function({
   });
 
   ipcMain.on('lndAtplInit', async event => {
-    atplProtoPath = path.join(__dirname, '..', 'assets', 'autopilot.proto');
-    const packageDef = protoLoader.loadSync(atplProtoPath, protoOptions);
+    const atplProto = path.join(__dirname, '..', 'assets', 'autopilot.proto');
+    const packageDef = protoLoader.loadSync(atplProto, protoOptions);
     let autopilotRpc = grpc.loadPackageDefinition(packageDef).autopilotrpc;
     autopilot = new autopilotRpc.Autopilot(`localhost:${lndPort}`, credentials);
     grpc.waitForClientReady(autopilot, Infinity, err => {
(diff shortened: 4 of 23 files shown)


.travis.yml
```diff
@@ -13,7 +13,7 @@ matrix:
 
 env:
   global:
-    - LND_TAG=d2186cc9da29853091175189268b073f49586cf0
+    - LND_TAG=3be894bc2857b319a9978a2b7de11cc2c427d316
```
tanx: The master should be on a newer version. Could you rebase on that one?

assets/autopilot.proto
```diff
@@ -0,0 +1,75 @@
...
+    string heuristic = 1 [json_name = "heuristic"];
+
+    /**
+    A map from hex-encoded public keys to scores. Scores must be in the range
+    [0.0, 1.0].
+    */
+    map<string, double> scores = 2 [json_name = "scores"];
+}
+
+message SetScoresResponse {}
```
tanx: Is there a reason why the autopilot proto isn't in the same proto file? The unlocker and ln rpcs are also in a single file.

public/grpc-client.js
```diff
@@ -115,6 +146,14 @@ module.exports.init = async function({
...
   });
 
+  ipcMain.on('autopilotRequest', (event, { method, body = {} }) => {
+    const deadline = new Date(new Date().getTime() + GRPC_TIMEOUT);
+    const handleResponse = (err, response) => {
+      event.sender.send(`autopilotResponse_${method}`, { err, response });
+    };
+    autopilot[method](body, { deadline }, handleResponse);
+  });
+
```
tanx: Seems like a pretty diff for just toggling autopilot. What's the advantage over restarting lnd with autopilot args disabled?

assets/autopilot.proto
valentinewallace: I could manually concatenate them into a single file, but in `lnd` `autopilot.proto` is its own separate file.

public/grpc-client.js
valentinewallace: If we restart `lnd`, the user will have to input their password again.

It's a pretty big diff, but we'll need most of the diff (the new protos, new autopilot methods, etc)  to implement the bos score feature anyway.

public/preload.js
```diff
@@ -6,7 +6,9 @@ const _ipcRenderer = require('electron').ipcRenderer;
 
 const filter = event => {
   if (
-    !/^(lnd)|(unlock)|(log)|(locale)|(open-url)[a-zA-Z_-]{0,20}$/.test(event)
+    !/^(lnd)|(unlock)|(log)|(locale)|(autopilot)|(open-url)[a-zA-Z_-]{0,20}$/.test(
+      event
+    )
```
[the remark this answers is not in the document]
valentinewallace: whoops, thanks! forgot.

package.json
```diff
@@ -26,7 +26,7 @@
     "build-storybook": "build-storybook -s public",
     "build-icon": "svgr --native --icon --filename-case kebab -d src/asset/icon src/asset/icon",
     "build-img": "svgr --native --filename-case kebab -d src/asset/img src/asset/img",
-    "protobuf": "pbjs assets/rpc.proto -t static-module -w es6 -o assets/rpc.js"
+    "protobuf": "pbjs assets/rpc.proto -t static-module -w es6 -o assets/rpc.js && pbjs assets/autopilot.proto -t static-module -w es6 -o assets/autopilot.js"
```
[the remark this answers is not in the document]
valentinewallace: removed

public/grpc-client.js
```diff
@@ -94,6 +97,34 @@ module.exports.init = async function({
...
+      'assets',
+      'autopilot.proto'
+    );
+    const options = {
+      keepCase: false,
+      longs: Number,
+      enums: String,
+      defaults: true,
+      oneofs: true,
+    };
```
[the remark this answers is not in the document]
valentinewallace: fixed

assets/autopilot.proto
tanx: @Roasbeef ?

tanx: It just makes this PR so much more complex if we need to fire up yet another grpc client for a new api.

public/grpc-client.js
```diff
@@ -114,6 +133,14 @@ module.exports.init = async function({
     lnd[method](body, { deadline }, handleResponse);
   });
 
+  ipcMain.on('lndAtplRequest', (event, { method, body = {} }) => {
+    const deadline = new Date(new Date().getTime() + GRPC_TIMEOUT);
```
tanx: I don't actually think we use these `deadline` dates anywhere. We can remove them from all the calls ... also above for unlocker/lnd events.

public/grpc-client.js
```diff
@@ -52,23 +52,26 @@ module.exports.init = async function({
   lndSettingsDir,
   network,
 }) {
+  const grpcOptions = {
+    keepCase: false,
+    longs: Number,
+    enums: String,
+    defaults: true,
+    oneofs: true,
+  };
```
tanx: Rename to `protoOptions`.

public/grpc-client.js
```diff
@@ -93,6 +96,22 @@ module.exports.init = async function({
     });
   });
 
+  ipcMain.on('lndAtplInit', async event => {
+    credentials = await getCredentials(lndSettingsDir);
```
tanx: Any reason we need to reload the credentials? Why not use the one already loaded above?

public/grpc-client.js
```diff
@@ -93,6 +96,22 @@ module.exports.init = async function({
...
+    const macaroonCreds = getMacaroonCreds(lndSettingsDir, network);
+    credentials = grpc.credentials.combineChannelCredentials(
+      credentials,
+      macaroonCreds
+    );
```
tanx: Same here ... have you tried reusing the credentials object combined in the lndEvent?

src/action/grpc.js
```diff
@@ -64,6 +64,17 @@ class GrpcAction {
...
 
+  /**
+   * This is called to initialize the GRPC client to autopilot. Once `autopilotReady`
+   * is set to true on the store GRPC calls can be made to the client.
+   * @return {Promise<undefined>}
+   */
+  async initAutopilot() {
+    await this._sendIpc('lndAtplInit', 'lndAtplReady');
+    log.info('GRPC autopilotReady');
+    this._store.autopilotReady = true;
```
tanx: Didn't see a place where we use `store.autopilotReady`. We can probably remove it from the store until we do.

public/grpc-client.js
```diff
@@ -52,23 +52,26 @@ module.exports.init = async function({
...
+    defaults: true,
+    oneofs: true,
+  };
   let credentials;
   let protoPath;
+  let atplProtoPath;
   let lnrpc;
   let unlocker;
   let lnd;
+  let autopilotRpc;
```
tanx: Declare these two `let` vars in the event handler if they're only used there.

src/action/grpc.js
valentinewallace: Ah sorry I didn't add this previously but we need it here https://github.com/lightninglabs/lightning-app/pull/1000/files#diff-24b8ff244eb737acf60f151b492a3034R90 to ensure the user doesn't toggle autopilot before the service is ready, and it's also needed for bos scores since otherwise they'll attempt to `SetScores` before the autopilot service is ready.

public/grpc-client.js
valentinewallace: fixed

public/grpc-client.js
valentinewallace: fixed

public/grpc-client.js
valentinewallace: hm true, I thought there was potential that autopilot would attempt to use the credentials before they're set properly [here](https://github.com/lightninglabs/lightning-app/blob/master/public/grpc-client.js#L86) but can't seem to repro that, so fixed :woman_shrugging:

public/grpc-client.js
valentinewallace: removed this declaration. the other one is used elsewhere, so I left it.

valentinewallace:
@tanx I realized that it would also be good to remember the user's settings and start `lnd` with autopilot deactivated if they previously set that in settings, so I added that feature as well.

src/action/grpc.js
tanx: Ah yes good point.

test/integration/action/action-integration.spec.js
```diff
@@ -228,6 +228,27 @@ describe('Action Integration Tests', function() {
...
+      status = await grpc1.sendAutopilotCommand('status');
+      expect(status.active, 'to be false');
+
+      status = await grpc2.sendAutopilotCommand('status');
+      expect(status.active, 'to be true');
+      await grpc2.sendAutopilotCommand('modifyStatus', { enable: false });
+      status = await grpc2.sendAutopilotCommand('status');
+      expect(status.active, 'to be false');
+    });
+
```
tanx: Nice that we have integration tests. But we should test the higher level `setting.toggleAutopilot()` action instead of the grpc and check that `store.settings.autopilot` is set correctly. We should also not forget to test turning autopilot on again.

src/action/grpc.js
```diff
@@ -64,6 +64,26 @@ class GrpcAction {
...
+    if (this._store.settings.autopilot) {
+      try {
+        await this.sendAutopilotCommand('modifyStatus', {
+          enable: true,
+        });
+      } catch (err) {
+        log.error('Failed to activate autopilot', err);
+      }
+    }
+    this._store.autopilotReady = true;
```
tanx: I'm thinking we should move `grpc.initAutopilot()` and `setting.toggleAutopilot()` into its own `autopilot.js` action module. Feels like we're creating duplicate try/catch code for both here and we should create a common wrapper around the grpc call.

public/lnd-child-process.js
```diff
@@ -76,6 +76,7 @@ module.exports.startLndProcess = async function({
     ]);
   } else {
     args = args.concat([
+      '--autopilot.active',
```
valentinewallace: In this [comment](https://github.com/lightninglabs/lightning-app/pull/1000/commits/f577a413ad1f2db3aaeaeefeadc73d23410f2e57#r266288726) you said you removed the flag, but it seems to be re-added here?

Note it will need to be removed for bos score :)

README.md
```diff
@@ -27,7 +27,7 @@ See the `ToDo (next release)` column on our [project board](https://github.com/l
 ```
 git clone https://github.com/lightningnetwork/lnd $GOPATH/src/github.com/lightningnetwork/lnd
 cd $GOPATH/src/github.com/lightningnetwork/lnd
-make && make install tags=experimental
+make && make install tags="experimental autopilotrpc"
```
valentinewallace: Good catch.. There are a ton of places this flag update is needed, hard to get them all :sweat_smile:

test/integration/action/action-integration.spec.js
valentinewallace: Are you saying remove [these](https://github.com/lightninglabs/lightning-app/pull/1000/files#diff-a2c88e3a85c1882dbd0b41495eb2b965R238) tests and add testing for autopilot-on to [these](https://github.com/lightninglabs/lightning-app/pull/1000/files#diff-226a70f557ce7896930bef43709bcd75R100) tests?

README.md
```diff
@@ -75,7 +75,7 @@ To run the packaged version of the app e.g. for macOS run:
 
 The app is configured for testnet by default but you can opt-in to mainnet if you have a btcd node running. Be aware that this is currently still experimental:
 ```
-./dist/mac/Lightning.app/Contents/MacOS/Lightning --bitcoin.mainnet --bitcoin.node=neutrino --neutrino.connect=127.0.0.1 --autopilot.active
+./dist/mac/Lightning.app/Contents/MacOS/Lightning --bitcoin.mainnet --bitcoin.node=neutrino --neutrino.connect=127.0.0.1
```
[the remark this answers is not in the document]
valentinewallace: sounds good

test/integration/action/action-integration.spec.js
valentinewallace: fixed

src/action/grpc.js
valentinewallace: fixed

src/action/index.js
```diff
@@ -86,8 +89,9 @@ when(
  * lnd node all balances, channels and transactions are fetched.
  */
 when(
-  () => store.lndReady,
+  () => store.lndReady && store.autopilotReady,
   () => {
+    autopilot.init();
```
tanx: @valentinewallace I also split up init to be in line with our other actions.

src/action/grpc.js
```diff
@@ -49,6 +49,31 @@ class GrpcAction {
...
+  /**
+   * Wrapper function to execute calls to the autopilot grpc client.
+   * @param  {string} method The autopilot GRPC api to call
+   * @param  {Object} body   The payload passed to the api
+   * @return {Promise<Object>}
+   */
+  async sendAutopilotCommand(method, body) {
+    return this._sendIpc('lndAtplRequest', 'lndAtplResponse', method, body);
+  }
+
```
tanx: @valentinewallace I also moved these back to the grpc module since there was quite a bit of copy/paste ;)

src/component/setting.js
```diff
@@ -115,7 +115,7 @@ export const SettingItem = ({ name, onSelect, label, arrow, children }) => (
 
 SettingItem.propTypes = {
   name: PropTypes.string.isRequired,
-  onSelect: PropTypes.func.isRequired,
+  onSelect: PropTypes.func,
```
tanx: This event handler is no longer required. Otherwise you could click the item and nothing happened.

test/integration/action/action-integration.spec.js
```diff
@@ -345,6 +350,34 @@ describe('Action Integration Tests', function() {
...
+      let status = await grpc1.sendAutopilotCommand('status');
+      expect(status.active, 'to be false');
+
+      await autopilot1.toggle();
+      expect(store1.settings.autopilot, 'to be true');
+      status = await grpc1.sendAutopilotCommand('status');
+      expect(status.active, 'to be true');
+    });
+  });
+
```
tanx: Also refactored the integration tests and wrapped them into their own describe block.

test/unit/action/grpc.spec.js
```diff
@@ -81,6 +81,14 @@ describe('Action GRPC Unit Tests', () => {
     });
   });
 
+  describe('initLnd()', () => {
```
valentinewallace: Good point :sweat_smile:

test/integration/action/action-integration.spec.js
```diff
@@ -176,6 +180,7 @@ describe('Action Integration Tests', function() {
     channels2 = new ChannelAction(store2, grpc2, nav2, notify2);
     invoice2 = new InvoiceAction(store2, grpc2, nav2, notify2);
     payments2 = new PaymentAction(store2, grpc2, nav2, notify2);
+    autopilot2 = new AtplAction(store2, grpc2, db2, notify2);
```
valentinewallace: I know uniformity is good, but wanted to point out this isn't used anywhere. I think that's alright though.

PR #1005, opened 2019-03-12
Bos scores

### LINKED ISSUE #986: Bos score api integration for testnet/mainnet

valentinewallace:
Closes #986. Depends on #1000.


valentinewallace:
> @valentinewallace can we create a PR against master with just the bos score stuff (separately from autopilot toggle)? Or do we need the new grpc/proto apis?

We need the new grpc/proto apis to call `SetScores`.

src/action/channel.js
```diff
@@ -241,7 +241,7 @@ class ChannelAction {
     try {
       const networkStr =
         (await this._getNetwork()) == 'testnet' ? 'testnet' : '';
-      const uri = `http://nodes.lightning.computer/availability/btc${networkStr}.json`;
+      const uri = `https://nodes.lightning.computer/availability/btc${networkStr}.json`;
```
tanx: 👍

public/lnd-child-process.js
```diff
@@ -65,6 +65,8 @@ module.exports.startLndProcess = async function({
     '--autopilot.private',
     '--autopilot.minconfs=0',
     '--autopilot.allocation=0.95',
+    '--autopilot.heuristic=externalscore:0.7',
+    '--autopilot.heuristic=preferential:0.3',
```
tanx: @alexbosworth what do these numbers mean? Do these values make sense a a preset?

src/action/autopilot.js
```diff
@@ -14,12 +18,15 @@ class AtplAction {
   /**
    * Initialize autopilot from the stored settings and enable it via grpc
    * depending on if the user has enabled it in the last session.
+   * Additionally, poll the bos scores either way.
    * @return {Promise<undefined>}
    */
   async init() {
+    await this.updateAutopilotScores();
```
tanx: We don't need this initial call since `poll` will sleep after the first poll.

src/action/autopilot.js
```diff
@@ -48,6 +55,69 @@ class AtplAction {
...
+
+  /**
+   * Fetch bos scores, persist them, and return them in a list.
+   * @return {Object}
+   */
+  async fetchBosScores() {
+    const networkStr =
+      (await this._getNetwork()) === 'testnet' ? 'testnet' : '';
+    try {
+      const uri = `https://nodes.lightning.computer/availability/btc${networkStr}.json`;
```
tanx: We should move the base url `https://nodes.lightning.computer` to config.js

src/store.js
```diff
@@ -91,6 +91,8 @@ export class Store {
         exchangeRate: {},
         restoring: false,
         autopilot: true,
+        atplScores: {},
+        atplTestnetScores: {},
```
tanx: Rename to `atplScoresTestnet`

src/action/autopilot.js
```diff
@@ -48,6 +55,69 @@ class AtplAction {
...
+      let score = node.score / 100000000.0;
+      scores[node.public_key] = score;
+    }
+    return scores;
+  }
+
+  async _getNetwork() {
+    const response = await this._grpc.sendCommand('getInfo');
+    return response.chains[0].network;
+  }
```
tanx: We should set `store.network` in the info action when we poll `getInfo`. No need to do an extra grpc roundtrip here.

src/action/autopilot.js
tanx: Oh and it's
mainnet: https://nodes.lightning.computer/availability/btc.json
testnet: https://nodes.lightning.computer/availability/btctestnet.json

public/lnd-child-process.js
```diff
@@ -65,6 +65,8 @@ module.exports.startLndProcess = async function({
     '--autopilot.private',
     '--autopilot.minconfs=0',
     '--autopilot.allocation=0.95',
+    '--autopilot.heuristic=externalscore:1',
+    '--autopilot.heuristic=preferential:0',
```
tanx: I'd like to rely 100% on external scores for now to test how reliable/deterministic the UX is.

src/action/autopilot.js
```diff
@@ -48,6 +54,62 @@ class AtplAction {
...
+   */
+  async updateNodeScores() {
+    try {
+      await this._setNetwork();
+      const scores = await this._readNodeScores();
+      await this._setNodeScores(scores);
+    } catch (err) {
+      log.error('Updating autopilot scores failed', err);
+    }
+  }
```
tanx: I split the high level api into smaller functions. This makes the code much more readable than putting everything in a single try/catch block with nested if/else statements. We should try to split actions up like this for better readability.

test/unit/action/autopilot.spec.js
```diff
@@ -59,4 +70,97 @@ describe('Action Autopilot Unit Test', () => {
...
+      expect(grpc.sendAutopilotCommand, 'was called with', 'setScores', {
+        heuristic: 'externalscore',
+        scores: { 'some-pubkey': 0.14035087 },
+      });
+      expect(db.save, 'was called once');
+      expect(store.settings.nodeScores, 'to equal', {
+        mainnet: { 'some-pubkey': 0.14035087 },
+      });
+    });
+  });
```
tanx: We should also test all relevant code branches much more thoroughly. I rewrote the unit tests from scratch.

src/action/autopilot.js
```diff
@@ -3,20 +3,26 @@
...
 
   /**
    * Initialize autopilot from the stored settings and enable it via grpc
-   * depending on if the user has enabled it in the last session.
+   * depending on if the user has enabled it in the last session. Fetch node
+   * scores are fetched from an api to inform channel selection.
    * @return {Promise<undefined>}
    */
   async init() {
+    await this.updateNodeScores();
```
valentinewallace: I believe we're supposed to be polling the bos scores every 1-2 hours to ensure they're up to date.

src/action/autopilot.js
```diff
@@ -48,6 +54,62 @@ class AtplAction {
...
+
+  _formatNodesScores(jsonScores) {
+    return jsonScores.reduce((map, node) => {
+      map[node.public_key] = node.score / 100000000.0;
+      return map;
+    }, {});
+  }
+
+  async _setNodeScores(scores) {
+    if (!scores) {
```
valentinewallace: IIUC because `{}` is truthy, we'll want to change this to `if (scores === {}) {`.

test/unit/action/autopilot.spec.js
```diff
@@ -2,14 +2,18 @@ import { Store } from '../../../src/store';
 import GrpcAction from '../../../src/action/grpc';
 import NotificationAction from '../../../src/action/notification';
 import AppStorage from '../../../src/action/app-storage';
+import InfoAction from '../../../src/action/info';
 import AtplAction from '../../../src/action/autopilot';
 import * as logger from '../../../src/action/log';
+import nock from 'nock';
+import 'isomorphic-fetch';
```
valentinewallace: oo fancy

src/action/autopilot.js
tanx: Yeah. How often are the scores updated though in reality? The current solution updates them everytime the user starts the app, which is I think good enough for the beginning. We can add polling in the future if we really need it.

src/action/autopilot.js
tanx: scores is `store.settings.nodeScores[network]`. This will be undefined if not set since  `store.settings.nodeScores` is initialized as `{}` now.

tanx:
@valentinewallace I added polling of scores again every hour. I also added validation of the scores json format in case we change the data format for some reason.

PR #1008, opened 2019-03-12
Display blocks to maturity on channel detail view.

### LINKED ISSUE #525: Show days+hours for force closed channels in channel detail view
Atm for force closed channels we show this on the main channel page
screen shot 2018-08-13 at 1 02 03 pm

As the channel is no longer _active_ and can't be used for payments, we can probably blank out the can send/receive columns. 

When users click down in they get this: 
screen shot 2018-08-13 at 5 03 39 pm

As is the ID field is blank, so users are unable to lookup the txid of the closing transaction on chain. Note that this is different from the txid of the _funding_ transaction. We can get the _closing_ txid from the `ForceClosedChannel` proto: https://api.lightning.community/#forceclosedchannel

From the same proto, we can also obtain the "blocks until maturity". Atm we don't display any sort of time estimate to when the funds will be swept back into the wallet. On mainnet, this can be anxiety inducing as you don't know when the funds will be swept back into your wallet. We've received a ton of issues just related to users not quite understanding timelocks on `lnd`, which caused us to surface more info on the command line. 

To start with, we can either display the block of blocks (which at least counts down), or attempt to map the # of blocks to days+hours. One block is 10 mins, and a days worth of blocks is 144, so we can use that as a starting point from there. The goal here to give users this feedback in "dark mode", so we can set proper expectations w.r.t _when_ their funds will be restored.

valentinewallace:
Closes #525.

"Blocks to maturity" doesn't seem like the best copy. Any ideas for what would be better?

### DIFF
--- a/src/action/channel.js
+++ b/src/action/channel.js
@@ -3,7 +3,7 @@
  * call the corresponding GRPC apis for channel management.
  */
 
-import { toSatoshis, poll, formatTimeTilMaturity } from '../helper';
+import { toSatoshis, poll, getTimeTilAvailable } from '../helper';
 import * as log from './log';
 
 class ChannelAction {
@@ -160,7 +160,7 @@ class ChannelAction {
         limboBalance: pfcc.limboBalance,
         maturityHeight: pfcc.maturityHeight,
         blocksTilMaturity: pfcc.blocksTilMaturity,
-        timeTilAvailable: formatTimeTilMaturity(pfcc.blocksTilMaturity),
+        timeTilAvailable: getTimeTilAvailable(pfcc.blocksTilMaturity),
         status: 'pending-force-closing',
       }));
       const wccs = response.waitingCloseChannels.map(wcc => ({
--- a/src/action/channel.js
+++ b/src/action/channel.js
@@ -3,7 +3,7 @@
  * call the corresponding GRPC apis for channel management.
  */
 
-import { toSatoshis, poll } from '../helper';
+import { toSatoshis, poll, formatTimeTilMaturity } from '../helper';
 import * as log from './log';
 
 class ChannelAction {
@@ -160,7 +160,7 @@ class ChannelAction {
         limboBalance: pfcc.limboBalance,
         maturityHeight: pfcc.maturityHeight,
         blocksTilMaturity: pfcc.blocksTilMaturity,
-        timeTilAvailable: this._formatTimeTilMaturity(pfcc.blocksTilMaturity),
+        timeTilAvailable: formatTimeTilMaturity(pfcc.blocksTilMaturity),
         status: 'pending-force-closing',
       }));
       const wccs = response.waitingCloseChannels.map(wcc => ({
@@ -346,12 +346,6 @@ class ChannelAction {
     const pc = this._store.pendingChannels;
     const channel = pc.find(c => c.channelPoint === channelPoint);
     if (channel) pc.splice(pc.indexOf(channel));
-  }
-
-  _formatTimeTilMaturity(numBlocks) {
-    const days = Math.floor(numBlocks / (24 * 6));
-    const hours = Math.floor((numBlocks % (24 * 6)) / 6);
-    return `${days} days and ${hours} hours`;
   }
 }
 
--- a/src/action/channel.js
+++ b/src/action/channel.js
@@ -160,6 +160,7 @@ class ChannelAction {
         limboBalance: pfcc.limboBalance,
         maturityHeight: pfcc.maturityHeight,
         blocksTilMaturity: pfcc.blocksTilMaturity,
+        timeTilAvailable: this._formatTimeTilMaturity(pfcc.blocksTilMaturity),
         status: 'pending-force-closing',
       }));
       const wccs = response.waitingCloseChannels.map(wcc => ({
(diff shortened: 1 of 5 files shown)


src/view/channel-detail.js
```diff
@@ -42,6 +42,11 @@ const ChannelDetailView = ({ store, nav }) => (
             {store.selectedChannel.closingTxId}
           </DetailField>
         ) : null}
+        {store.selectedChannel.blocksTilMaturity ? (
+          <DetailField name="Blocks Until Maturity">
```
Roasbeef: Rather than showing blocks, I think it'd be more intuitive for end users to show days/hours. So we can convert the blocks to days/hours using 144 blocks in a day (6 each hour).

Roasbeef: As for the copy, maybe: "Time Til Available", "Time Til Recovery"?

valentinewallace: fixed

src/action/channel.js
```diff
@@ -346,6 +347,12 @@ class ChannelAction {
     const channel = pc.find(c => c.channelPoint === channelPoint);
     if (channel) pc.splice(pc.indexOf(channel));
   }
+
+  _formatTimeTilMaturity(numBlocks) {
+    const days = Math.floor(numBlocks / (24 * 6));
+    const hours = Math.floor((numBlocks % (24 * 6)) / 6);
+    return `${days} days and ${hours} hours`;
+  }
```
tanx: Probably better to move to helper.js for easier unit testing

valentinewallace: fixed

src/action/channel.js
```diff
@@ -160,6 +160,7 @@ class ChannelAction {
         limboBalance: pfcc.limboBalance,
         maturityHeight: pfcc.maturityHeight,
         blocksTilMaturity: pfcc.blocksTilMaturity,
+        timeTilAvailable: formatTimeTilMaturity(pfcc.blocksTilMaturity),
```
tanx: Rename to `getTimeTilAvailable` ... we're getting the time from the blocks.

src/view/channel-detail.js
```diff
@@ -42,6 +42,11 @@ const ChannelDetailView = ({ store, nav }) => (
             {store.selectedChannel.closingTxId}
           </DetailField>
         ) : null}
+        {store.selectedChannel.blocksTilMaturity ? (
```
tanx: This will always be falsy ... change to `store.selectedChannel.timeTilAvailable`

test/unit/helper.spec.js
```diff
@@ -817,4 +817,12 @@ describe('Helpers Unit Tests', () => {
...
     });
   });
+
+  describe('formatTimeTilMaturity()', () => {
+    it('should format blocks to human-readable time', () => {
+      const numBlocks = 463;
+      const time = helpers.formatTimeTilMaturity(numBlocks);
+      expect(time, 'to equal', '3 days and 5 hours');
+    });
+  });
```
tanx: Nice test. Maybe add test cases for edge values like `undefined` and `0`.

src/action/channel.js
valentinewallace: fixed

src/view/channel-detail.js
valentinewallace: fixed

test/unit/helper.spec.js
valentinewallace: fixed

src/view/channel-detail.js
```diff
@@ -42,6 +42,11 @@ const ChannelDetailView = ({ store, nav }) => (
             {store.selectedChannel.closingTxId}
           </DetailField>
         ) : null}
+        {store.selectedChannel.timeTilAvailable ? (
+          <DetailField name="Time Til Available">
+            {store.selectedChannel.timeTilAvailable}
+          </DetailField>
+        ) : null}
```
tanx: Move this below `statusLabel`. Seems more natural to reading the time til available after the status.

test/unit/helper.spec.js
```diff
@@ -817,4 +817,29 @@ describe('Helpers Unit Tests', () => {
...
+  describe('getTimeTilAvailable()', () => {
+    it('should format blocks to human-readable time', () => {
+      const numBlocks = 463;
+      const time = helpers.getTimeTilAvailable(numBlocks);
+      expect(time, 'to equal', '3 days and 5 hours');
+    });
+
+    it('should show NaN for undefined', () => {
+      const time = helpers.getTimeTilAvailable(undefined);
+      expect(time, 'to equal', 'NaN days and NaN hours');
```
tanx: We should probably throw an Error for `undefined`. This should only happen if the api changed and `blocksTilMaturity` isn't set as expected. And we wanna get notified in that case.

valentinewallace: fixed

src/view/channel-detail.js
valentinewallace: fixed

src/helper.js
```diff
@@ -321,3 +321,21 @@ export const generateArc = (x, y, radius, startAngle, endAngle) => {
...
 };
+
+/**
+ * Convert a number of blocks to an amount of time in the format "X days and Y
+ * hours" assuming 10 minutes per block.
+ * @param  {number} numBlocks The number of blocks to convert.
+ * @return {string} The amount of time the blocks is equivalent to.
+ */
+export const getTimeTilAvailable = numBlocks => {
+  let num = Number(numBlocks);
```
tanx: This casting should not be necessary since `blocks_til_maturity` is of type number.

PR #1009, opened 2019-03-13
Include limbo balance in Home screen balance.

### LINKED ISSUE #1006: Add total_limbo_balance to unified balance on home screen
The `total_limbo_balance` field from pending channels should be added to the unified balance on the home screen balance sum: https://api.lightning.community/#pendingchannels

We should also add it to the `Closing` sum at the top of the channel page on desktop.

Otherwise, if the user has only one channel, and they force close it for w/e reason, their balance on the home screen will go to zero.

valentinewallace:
Closes #1006.

### DIFF
--- a/src/action/wallet.js
+++ b/src/action/wallet.js
@@ -144,7 +144,11 @@ class WalletAction {
    * @return {Promise<undefined>}
    */
   async update() {
-    await Promise.all([this.getBalance(), this.getChannelBalance()]);
+    await Promise.all([
+      this.getBalance(),
+      this.getChannelBalance(),
+      this.getLimboBalance(),
+    ]);
   }
 
   /**
@@ -446,6 +450,20 @@ class WalletAction {
       this._store.pendingBalanceSatoshis = r.pendingOpenBalance;
     } catch (err) {
       log.error('Getting channel balance failed', err);
+    }
+  }
+
+  /**
+   * Fetch the limbo balance using the lnd grpc api and set the
+   * corresponding value on the global store.
+   * @return {Promise<undefined>}
+   */
+  async getLimboBalance() {
+    try {
+      const r = await this._grpc.sendCommand('PendingChannels');
+      this._store.limboBalanceSatoshis = r.totalLimboBalance;
+    } catch (err) {
+      log.error('Getting limbo balance failed', err);
     }
   }
 
--- a/src/action/wallet.js
+++ b/src/action/wallet.js
@@ -144,11 +144,7 @@ class WalletAction {
    * @return {Promise<undefined>}
    */
   async update() {
-    await Promise.all([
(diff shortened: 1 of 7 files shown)


src/action/wallet.js
```diff
@@ -449,6 +453,20 @@ class WalletAction {
...
+   */
+  async getLimboBalance() {
+    try {
+      const r = await this._grpc.sendCommand('PendingChannels');
+      this._store.limboBalanceSatoshis = r.totalLimboBalance;
+    } catch (err) {
+      log.error('Getting limbo balance failed', err);
+    }
+  }
+
```
tanx: `store.limboBalanceSatoshis` should be done in `getPendingChannels` of channel actions as that grpc api is already being polled. Doing it here would result in the api being polled twice.

src/computed/wallet.js
```diff
@@ -21,8 +21,14 @@ const ComputedWallet = store => {
...
         channelBalanceSatoshis,
+        limboBalanceSatoshis,
       } = store;
-      return balanceSatoshis + pendingBalanceSatoshis + channelBalanceSatoshis;
+      return (
+        balanceSatoshis +
+        pendingBalanceSatoshis +
+        channelBalanceSatoshis +
+        limboBalanceSatoshis
+      );
```
tanx: Although `store.limboBalanceSatoshis` is set in channel actions I would also compute this here 👍

test/unit/computed/wallet.spec.js
```diff
@@ -51,10 +52,11 @@ describe('Computed Wallet Unit Tests', () => {
       store.balanceSatoshis = 50000001;
       store.pendingBalanceSatoshis = 50000000;
       store.channelBalanceSatoshis = 10000;
+      store.limboBalanceSatoshis = 100;
       store.settings.unit = 'bit';
       ComputedWallet(store);
-      expect(store.totalBalanceSatoshis, 'to equal', 100010001);
-      expect(store.totalBalanceLabel, 'to match', /1[,.]000[,.]100[,.]01/);
+      expect(store.totalBalanceSatoshis, 'to equal', 100010101);
+      expect(store.totalBalanceLabel, 'to match', /1[,.]000[,.]101[,.]01/);
```
tanx: Nice test 👍

src/computed/wallet.js
valentinewallace: Sorry, I'm confused what this means. What is being computed?

src/action/wallet.js
valentinewallace: fixed

src/computed/wallet.js
tanx: Just saying it makes sense to use the `ComputedWallet` module here. So no changes needed :)

PR #1019, opened 2019-03-15
Fix balances

### LINKED ISSUE #1007: Add total_limbo_balance to "Closing" sum on channels page
We should add `total_limbo_balance` it to the Closing sum at the top of the channel page on desktop: https://api.lightning.community/#pendingchannels

### LINKED ISSUE #1016: Pending-open balance is being added to unifued balance twice
We need to remove `pendingBalanceSatoshis` from ComputedWallet.totalBalanceSatoshis()

https://github.com/lightninglabs/lightning-app/blob/4c586796ca9bf4ee2cbb585178fa81cb252691dd/src/computed/wallet.js#L19-L32

tanx:
Closes #1007, closes #1016

### DIFF
--- a/src/action/channel.js
+++ b/src/action/channel.js
@@ -168,7 +168,6 @@ class ChannelAction {
         status: 'waiting-close',
       }));
       this._store.pendingChannels = [].concat(pocs, pccs, pfccs, wccs);
-      this._store.limboBalanceSatoshis = response.totalLimboBalance;
     } catch (err) {
       log.error('Listing pending channels failed', err);
     }
--- a/src/computed/channel.js
+++ b/src/computed/channel.js
@@ -68,7 +68,7 @@ const ComputedChannel = store => {
     get channelBalanceClosingSatoshis() {
       return (store.pendingChannels || [])
         .filter(c => !c.status.includes('open'))
-        .map(c => Number(c.localBalance))
+        .map(c => c.localBalance)
         .reduce((a, b) => a + b, 0);
     },
     get channelBalanceClosingLabel() {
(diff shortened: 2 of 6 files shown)


src/action/channel.js
```diff
@@ -168,7 +168,6 @@ class ChannelAction {
         status: 'waiting-close',
       }));
       this._store.pendingChannels = [].concat(pocs, pccs, pfccs, wccs);
-      this._store.limboBalanceSatoshis = response.totalLimboBalance;
```
valentinewallace: Ah so [this](https://github.com/lightninglabs/lightning-app/issues/1007#issuecomment-472722390) comment is outdated? Basically we were calculating correctly to begin with? I forget why we initially thought `limboBalance` was relevant to anything :thinking:

src/computed/wallet.js
```diff
@@ -21,13 +21,13 @@ const ComputedWallet = store => {
         balanceSatoshis,
         pendingBalanceSatoshis,
         channelBalanceSatoshis,
-        limboBalanceSatoshis,
+        channelBalanceClosingSatoshis,
```
valentinewallace: Cool, this does make sense to fix the [double-add problem](https://github.com/lightninglabs/lightning-app/issues/1016#issuecomment-472725500). Between this and `pendingBalanceSatoshis` we cover all the pending funds, then `balanceSatoshis` is on-chain proper, and `channelBalanceSatoshis` is off-chain proper.

src/computed/channel.js
```diff
@@ -38,7 +38,7 @@ const ComputedChannel = store => {
     get channelBalanceOpenSatoshis() {
       return (store.channels || [])
         .filter(c => c.active)
-        .map(c => Number(c.localBalance))
+        .map(c => c.localBalance)
```
valentinewallace: Good catch :+1:

src/action/channel.js
tanx: Yes. They are one and the same. We can calculate limboBalanceSatoshis from the pendingChannels grpc response array.

PR #1025, opened 2019-03-19
Upgrade to go v1.12.1

tanx:


tanx:
> Travis seems OK with this change! Are you still doing macOS testing? I'll approve but with the caveat we should make sure we don't have a massive performance regression first ;D

@valentinewallace yeah. I'd like to figure out whether the lnd regression #1028 I discovered yesterday is caused by `go 1.12` or the current `lnd master`. If it's lnd master then the only thing holding `go 1.12` is the performance regression. As @halseth pointed out, it might make sense to prioritize disk integrity/safety over performance. We should still test a sync on testnet and mainnet with go 1.12 either way though to see if it really makes such a big difference to app users. If I understand correctly the performance regression is only be on desktop/macOS and as @Roasbeef pointed out, neutrino starts syncing instantly now anyway when the app is opened.

PR #1032, opened 2019-03-22
Show warning ntfn when generating an invoice that's too big to receive.

### LINKED ISSUE #1024: Display error message "not enough inbound capacity" when creating an invoice with a higher invoice amount

valentinewallace:
Closes #1024.

### DIFF
--- a/src/action/invoice.js
+++ b/src/action/invoice.js
@@ -84,11 +84,10 @@ class InvoiceAction {
    * @return {undefined}
    */
   checkAmount({ satAmount }) {
-    for (let i = 0; i < this._store.channels.length; i++) {
-      const chan = this._store.channels[i];
-      if (chan.remoteBalance >= satAmount) {
-        return;
-      }
+    const { channels } = this._store;
+    const hasInbound = channels.find(c => c.remoteBalance >= satAmount);
+    if (hasInbound) {
+      return;
     }
     this._notification.display({
       msg: "You don't have enough inbound capacity to receive this payment.",
--- a/src/action/invoice.js
+++ b/src/action/invoice.js
@@ -61,8 +61,10 @@ class InvoiceAction {
   async generateUri() {
     try {
       const { invoice, settings } = this._store;
+      const satAmount = toSatoshis(invoice.amount, settings);
+      this.checkAmount({ satAmount });
       const response = await this._grpc.sendCommand('addInvoice', {
-        value: toSatoshis(invoice.amount, settings),
+        value: satAmount,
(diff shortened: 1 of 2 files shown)


src/action/invoice.js
```diff
@@ -61,8 +61,10 @@ class InvoiceAction {
   async generateUri() {
     try {
       const { invoice, settings } = this._store;
+      const amount = toSatoshis(invoice.amount, settings);
+      this.checkAmount({ satAmount: amount });
```
tanx: We should do an early return in case `checkAmount()` returns false and there isn't enough inbound liquidity. Otherwise what's the value of displaying an invoice to the user?

test/unit/action/invoice.spec.js
```diff
@@ -77,7 +77,7 @@ describe('Action Invoice Unit Tests', () => {
       grpc.sendCommand.rejects(new Error('Boom!'));
       await invoice.generateUri();
       expect(nav.goInvoiceQR, 'was not called');
-      expect(notification.display, 'was called once');
+      expect(notification.display, 'was called twice');
     });
   });
```
tanx: We should add a test case for the early return (not enough inbound) case here.

src/action/invoice.js
valentinewallace: It might be fun for people who just want to "try out" the app :thinking:

tanx: That's true. I'm fine with just displaying a warning for now.

src/action/invoice.js
```diff
@@ -61,8 +61,10 @@ class InvoiceAction {
   async generateUri() {
     try {
       const { invoice, settings } = this._store;
+      const amount = toSatoshis(invoice.amount, settings);
```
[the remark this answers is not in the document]
valentinewallace: fixed

src/action/invoice.js
```diff
@@ -75,6 +77,24 @@ class InvoiceAction {
...
+  checkAmount({ satAmount }) {
+    for (let i = 0; i < this._store.channels.length; i++) {
+      const chan = this._store.channels[i];
+      if (chan.remoteBalance >= satAmount) {
+        return;
+      }
+    }
+    this._notification.display({
+      msg: "You don't have enough inbound capacity to receive this payment.",
+    });
```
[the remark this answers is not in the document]
valentinewallace: given [this](https://github.com/lightninglabs/lightning-app/pull/1032#discussion_r272002583) comment I skipped this part. lmk if I'm missing something.

PR #1034, opened 2019-03-23
Show tenths of a cent on transation and invoice confirmation views.

### LINKED ISSUE #1023: Show tenths of cents on invoice confirmation page
### Description

When entering an invoice for Yalls.org, the cost of an article is $0.006, but the App rounds it to $0.01. Would be useful to show tenths of a cent for very small invoices (e.g. < $0.05)

Create an invoice to view an article at Yalls.org (current cost is 150 sats, or $0.006)

### Expected behavior

Show the invoice value of $0.006.

### Actual behavior

Shows the invoice value of $0.01

valentinewallace:
Closes #1023.

This got me thinking we should show tenths of a cent everywhere... but maybe that's too off-putting for initial launch.

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -24,12 +24,13 @@ export const formatNumber = val => {
  * @param  {string} currency   The fiat currency e.g. 'usd'
  * @return {string}            The formatte fiat currency value
  */
-export const formatFiat = (val, currency) => {
+export const formatFiat = (val, currency, centTenths = false) => {
   let num = Number(val);
   if (isNaN(num)) {
     num = 0;
   }
-  const options = { style: 'currency', currency };
+  const minimumFractionDigits = centTenths && val % 1 !== 0 ? 3 : 2;
+  const options = { style: 'currency', minimumFractionDigits, currency };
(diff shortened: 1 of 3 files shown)


src/helper.js
```diff
@@ -113,15 +114,19 @@ export const calculateExchangeRate = (satoshis, settings) => {
...
   ) {
     throw new Error('Invalid input!');
   }
   return settings.displayFiat
-    ? formatFiat(calculateExchangeRate(satoshis, settings), settings.fiat)
+    ? formatFiat(
+        calculateExchangeRate(satoshis, settings),
+        settings.fiat,
+        centTenths
+      )
```
tanx: Couldn't we just change `formatFiat` to prepend a `< ` if the string matches `/0[,.]00/`?

valentinewallace: Do you mean if `formatFiat` is going to return `0.00`, it will instead return `<0.00`? I think I'm misunderstanding...

tanx: Exactly. Why do we need to add an arg to `toAmountLabel`?

tanx:
Closing in favor of #1049

PR #1035, opened 2019-03-27
Payscreen max toggle

### LINKED ISSUE #865: add a "max" toggle in the pay screen dialogue
Once this PR is merged into `lnd` it will be possible to instruct lnd to sweep all on chain coins to a single address. This can be useful if one wishes to switch wallets, or simply withdrawal all the funds in an existing wallet. The main [proto modification is the addition of a `send_all` flag](https://github.com/lightningnetwork/lnd/pull/2198/files#diff-f6ec8c0f52caccfaec5656028d316c7fR841). If this is set (and no amount is set) then we'll sweep all the coins into a single address. This was possible before, but only via trial and error as one had to guess the total fees (based on the number of UTXOs in the wallet) as the magic number to input for the send logic. Additionally, even if their guess was very close, they'd likely end up with a very small output that may not even be economical to spend as left over. 

We can integrate this new feature in the form of a "max" toggle in the pay screen dialogue. This is related to #863 as without that, if you sweep all the funds to the same address within lnd, then they'd all be used again to open channels.

valentinewallace:
Closes #865.

### DIFF
--- a/src/action/payment.js
+++ b/src/action/payment.js
@@ -122,6 +122,7 @@ class PaymentAction {
    */
   setAmount({ amount }) {
     this._store.payment.amount = amount;
+    this._store.payment.sendAll = false;
   }
 
   /**
@@ -139,6 +140,7 @@ class PaymentAction {
     if (await this.decodeInvoice({ invoice: this._store.payment.address })) {
       this._nav.goPayLightningConfirm();
     } else if (isAddress(this._store.payment.address)) {
+      this._store.payment.sendAll = false;
       this._nav.goPayBitcoin();
     } else {
       this._notification.display({ msg: 'Invalid invoice or address' });
@@ -238,11 +240,7 @@ class PaymentAction {
     }, PAYMENT_TIMEOUT);
     this._nav.goWait();
     try {
-      const { payment, settings } = this._store;
-      await this._grpc.sendCommand('sendCoins', {
-        addr: payment.address,
-        amount: toSatoshis(payment.amount, settings),
-      });
+      await this._sendPayment();
       this._nav.goPayBitcoinDone();
     } catch (err) {
       this._nav.goPayBitcoinConfirm();
@@ -250,6 +248,16 @@ class PaymentAction {
     } finally {
       clearTimeout(timeout);
     }
+  }
+
+  async _sendPayment() {
+    const { payment, settings } = this._store;
+    let amount = payment.sendAll ? 0 : toSatoshis(payment.amount, settings);
+    await this._grpc.sendCommand('sendCoins', {
+      addr: payment.address,
+      amount,
+      sendAll: payment.sendAll,
+    });
   }
 
   /**
--- a/src/action/payment.js
+++ b/src/action/payment.js
@@ -126,6 +126,37 @@ class PaymentAction {
   }
 
   /**
+   * Set the payment amount to the max amount that can be sent. This
+   * is useful for people to move their coins off of the app.
+   * @return {Promise<undefined>}
+   */
+  async setMax() {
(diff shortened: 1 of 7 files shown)


src/action/payment.js
```diff
@@ -122,6 +122,38 @@ class PaymentAction {
...
+   * is useful for people to move their coins off of the app.
+   * @return {Promise<undefined>}
+   */
+  async setMax() {
+    const timeout = setTimeout(() => {
+      return this._notification.display({
+        type: 'error',
+        msg: 'Setting max payment amount timed out!',
+      });
+    }, PAYMENT_TIMEOUT);
```
tanx: Replace with `await nap(PAYMENT_TIMEOUT)` from helpers.js

src/action/payment.js
```diff
@@ -122,6 +122,38 @@ class PaymentAction {
...
+        feeSat = await this._getFeeEstimateSat({
+          addr: payment.address,
+          amtSat,
+        });
+      } catch (err) {
+        amtSat = Math.floor(0.99 * amtSat);
+      } finally {
+        clearTimeout(timeout);
+      }
+    }
```
tanx: Do we really want to poll `_getFeeEstimateSat` like this? Can't we just display a static max amount which has the fee deducted? I mean the on chain fee shouldn't change much depending on the amount right? And even if it does change slightly it shouldn't be a big issues as long as all of the funds are flushed to an external wallet.

src/action/payment.js
```diff
@@ -199,12 +232,21 @@ class PaymentAction {
...
+
+  async _getFeeEstimateSat({ addr, amtSat }) {
     const AddrToAmount = {};
-    AddrToAmount[payment.address] = toSatoshis(payment.amount, settings);
+    AddrToAmount[addr] = amtSat;
     const { feeSat } = await this._grpc.sendCommand('estimateFee', {
       AddrToAmount,
     });
-    payment.fee = toAmount(feeSat, settings);
+    return feeSat;
```
tanx: Why do we need to create a new `_getFeeEstimateSat` helper function? Why not read `toSatoshis(payment.fee)`?

src/action/payment.js
```diff
@@ -252,6 +290,16 @@ class PaymentAction {
...
 
+  async _sendPayment() {
+    const { payment, settings } = this._store;
+    let amount = payment.sendAll ? 0 : toSatoshis(payment.amount, settings);
+    await this._grpc.sendCommand('sendCoins', {
+      addr: payment.address,
+      amount,
+      sendAll: payment.sendAll,
+    });
+  }
```
tanx: Nice pulling out a helper function here. Makes the `payBitcoin` more readable too.

src/action/payment.js
valentinewallace: Yeah, I will rework this part.

The reason I wrote it this way is because the `estimateFee` gRPC call fails if given the total wallet balance, saying the amount is too high. But rather than polling, I'll just find a balance reduction that works for all reasonable wallet balances.

tanx: What's the minimum sat amount hat can be sent in an on-chain tx? Couldn't we just query with that?

valentinewallace: In my testing I saw a fair amount of variation (once I saw $40 of fees) so I went with 98% of the total on-chain balance and that seems to work in the cases I tried.

src/action/payment.js
valentinewallace: lol fixed, yeah this was too complicated

src/action/payment.js
valentinewallace: n/a

PR #1039, opened 2019-04-01
move react-scripts to devDeps 

hackerrdave:
`react-scripts` is only needed for development work - moving this to devdeps helps to reduce prod module size and npm audit noise for production modules (react-scripts is responsible for most of the npm audit warnings)


tanx:
Thanks for the PR. I just tried this locally and `npm audit --production` still gives the warnings.

Closing this in favor of #1040

PR #1044, opened 2019-04-03
Poll new address

### LINKED ISSUE #1037: address on deposit page isn't refreshed after use
We still seem to cache the same address for an active instance of the wallet. Instead, we should be using the new `UNUSED_NESTED_PUBKEY_HASH = 3` type, and stop caching the wallet address internally. Instead, if [we pass a `3` here](https://github.com/lightninglabs/lightning-app/blob/5f76cbc4ae71c26ceb7d802510b84308c3662bb2/src/action/wallet.js#L480) then `lnd` will return the same address until it has been used on chain. As a result, the wallet can now spam this call without having to worry about address inflation.

tanx:
Closes #1037

### DIFF
--- a/src/action/index-mobile.js
+++ b/src/action/index-mobile.js
@@ -99,7 +99,6 @@ when(() => store.walletUnlocked, () => grpc.initLnd());
 when(
   () => store.lndReady,
   () => {
-    wallet.getNewAddress();
     wallet.pollBalances();
     wallet.pollExchangeRate();
     channel.pollChannels();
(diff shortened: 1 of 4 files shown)


tanx:
> One note is that the logs get spammed now with "newaddress" calls. But that seems like an acceptable tradeoff :)

@valentinewallace ah yes. @Roasbeef perhaps we can set those log entries to `debug` level just like we did for the `channellist` polling.

PR #1073, opened 2019-04-24
Remove getInfo call from autopilot action

### LINKED ISSUE #1072: Remove redundant getInfo calls during autopilot init
The notification screens show "Syncing complete" three times since we call `getInfo` during autopilot init to set the `store.network`. Since we now wait with invoking the autopilot grpc apis until after `store.syncedToChain` is true, we can safely remove the redundant calls.

tanx:
Closes #1072


tanx:
Closing in favor of #1073

PR #1080, opened 2019-04-26
Dev/notification dropdown

### LINKED ISSUE #1002: Notification bar slows down rendering during slide transition
We should use this component instead: https://github.com/testshallpass/react-native-dropdownalert

tanx:
Closes #1002

### DIFF
--- a/mobile/package-lock.json
+++ b/mobile/package-lock.json
@@ -9186,6 +9186,25 @@
         "public-encrypt": "^4.0.0"
       }
     },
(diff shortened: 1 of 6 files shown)


tanx:
> This is a big performance improvement for mobile notifications 💯 the difference at least on my phone is pretty night and day.

Yeah. It was unbearable on the iPhone XS since `SafeAreaView` computes view height dynamically.

PR #1082, opened 2019-04-26
Dev/commit fee

### LINKED ISSUE #1066: Add commit_fee sum from listchannels to home screen balance

tanx:
Closes #1066


valentinewallace:
I'm in favor of removing the unused `commitWeight` field, unless there is a reason to keep it around? I'm not sure what it even means. Otherwise, LGTM :1st_place_medal: :)

tanx:
> I'm in favor of removing the unused `commitWeight` field, unless there is a reason to keep it around? I'm not sure what it even means. Otherwise, LGTM 🥇 :)

Good call. Ok I removed `commitWeight` from the PR.

PR #1101, opened 2019-05-03
Add padding for notification dropdown on Android

tanx:


halseth:
Can you post a before/after screenshot? Don't think this affects the Pixel 2 XL.

src/component/notification-wrapper.js
```diff
@@ -18,6 +18,7 @@ const styles = StyleSheet.create({
   },
   dropdown: {
     backgroundColor: color.blackDark,
+    paddingTop: Platform.OS === 'android' ? 25 : 0,
```
halseth: I think this should be the height of the status bar instead: https://stackoverflow.com/questions/35436643/how-to-find-height-of-status-bar-in-android-through-react-native

tanx: @halseth good call. Didn't know about `StatusBar.currentHeight`. Fixed and tested on Pixel 2.

PR #1106, opened 2019-05-09
Dev/seed recovery mobile

### LINKED ISSUE #902: Enable seed recovery on mobile

valentinewallace:
Closes #902 and #729, and replaces #781.

This PR also includes changing desktop's seed recovery so that the user has to enter their password twice instead of once.


src/action/wallet.js
```diff
@@ -212,7 +220,10 @@ class WalletAction {
     }
     await this.initWallet({
       walletPassword: newPassword,
-      seedMnemonic: this._store.seedMnemonic.toJSON(),
+      recoveryWindow: this._store.settings.restoring ? RECOVERY_WINDOW : 0,
+      seedMnemonic: this._store.settings.restoring
+        ? this._store.wallet.restoreSeed.toJSON()
+        : this._store.seedMnemonic.toJSON(),
```
tanx: Why are we storing the same data in different places in the store?

src/action/wallet.js
```diff
@@ -337,6 +349,7 @@ class WalletAction {
   initPrevRestorePage() {
     if (this._store.wallet.restoreIndex >= 3) {
       this._store.wallet.restoreIndex -= 3;
+      this._store.wallet.focusedRestoreInd = this._store.wallet.restoreIndex;
```
tanx: If `store.wallet.focusedRestoreInd` can be computed from `this._store.wallet.restoreIndex` why is it stored twice?

src/view/main.js
```diff
@@ -83,9 +82,6 @@ class MainView extends Component {
         {route === 'RestoreSeed' && (
           <RestoreSeed store={store} wallet={wallet} />
         )}
-        {route === 'RestorePassword' && (
-          <RestorePassword store={store} wallet={wallet} nav={nav} />
-        )}
```
tanx: Nice 👍So we're using the same set password screens as during the "generate new wallet" setup now?

valentinewallace: yes indeed :)

src/action/wallet.js
valentinewallace: It can't be computed from `restoreIndex` all the time, because it's set [here](https://github.com/lightninglabs/lightning-app/pull/1106/files#diff-c52251eee698fdb351cf7e92c36e4136R79) and used to `autofocus` the next input when the user hits `Return` during restore. 

I tried to combine them but it makes [this](https://github.com/lightninglabs/lightning-app/pull/1106/files#diff-fbb9a2d7e2e2b04543043634746adfe1R334) logic complicated :( `restoreIndex` always needs to be the index of the first input of the screen, because it's [used](https://github.com/lightninglabs/lightning-app/blob/master/src/computed/seed.js#L23) to compute the current indexes to display.

Kind of hard to explain, I can send demo vids if this doesn't make sense.

src/action/wallet.js
valentinewallace: fixed

PR #1114, opened 2019-05-13
native android: add log emitting

halseth:
Emit logs from native Android bindings.


mobile/android/app/src/main/java/host/exp/exponent/LndNativeModule.java
```diff
@@ -158,6 +213,19 @@ public void run() {
         new Thread(startLnd).start();
     }
 
+
+    private void readToEnd(BufferedReader buf, boolean emit) throws IOException {
```
tanx: Nit. Extra new line here

halseth: Fixed.

PR #1120, opened 2019-05-15
Set StatusBar barStyle to „light-content“

### LINKED ISSUE #1118: Use LightContent status bar style on iOS
Currently the iOS status bar displays battery, time, bars, etc in a black font since it uses the default status bar style. Given that most of the text in the app is white, it would be great if the status bar was updated to use `UIStatusBarStyleLightContent` to unify the status bar with the rest of the UI :)

tanx:
Closes #1118


valentinewallace:
LGTM :raised_hands:

PR #1129, opened 2019-05-30
Trim seed word whitespace to avoid obscure errors.

### LINKED ISSUE #1127: Multiple usability UI UX issues when restoring a wallet
_(no real funds at this BIP39 seed phrase)_ 

Screenshot_2019-05-25_at_06_56_46

1. Typing words 3 at a time. Pressing enter. Focus should move to a new field

2. `endless` not on the list?

3. Clicking `<` to go back gets me to the homescreen, without ability to correct words

valentinewallace:
Closes #1127.

### DIFF
--- a/src/action/wallet.js
+++ b/src/action/wallet.js
@@ -53,7 +53,7 @@ class WalletAction {
    * @param {number} options.index The seed index
    */
   setRestoreSeed({ word, index }) {
-    this._store.seedMnemonic[index] = word;
+    this._store.seedMnemonic[index] = word.trim();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/action/wallet.js
```diff
@@ -53,7 +53,7 @@ class WalletAction {
    * @param {number} options.index The seed index
    */
   setRestoreSeed({ word, index }) {
-    this._store.seedMnemonic[index] = word;
+    this._store.seedMnemonic[index] = word.trim();
```
tanx: This is a really simple fix. A unit tests to reproduce/fix error states would be useful though.

valentinewallace: Added! Just realized I misnamed the branch as well.. next time will name `dev/trim-seed-words`.

tanx: Awesome!

PR #939, opened 2019-02-08
Reset pin ios

valentinewallace:
Closes #901.

PR #1130, opened 2019-06-01
Reset pin mobile

### LINKED ISSUE #901: Allow PIN change on mobile

valentinewallace:
Closes #901.

I reused the existing `Set Pin` screens and only made a new one for the `ResetPinSaved` view -- much smaller diff than the last PR.

### DIFF
--- a/src/action/auth-mobile.js
+++ b/src/action/auth-mobile.js
@@ -75,6 +75,8 @@ class AuthAction {
     const { auth } = this._store;
     if (auth[param]) {
       auth[param] = auth[param].slice(0, -1);
+    } else if (param === 'newPin' && this._store.walletUnlocked) {
+      this._nav.goSettings();
     } else if (param === 'pinVerify') {
       this.initSetPin();
     }
@@ -94,7 +96,11 @@ class AuthAction {
       return;
     }
     await this._setToKeyStore(PIN, newPin);
-    await this._generateWalletPassword();
+    if (this._store.walletUnlocked) {
+      this._nav.goResetPasswordSaved();
+    } else {
+      await this._generateWalletPassword();
+    }
   }
 
   /**
--- a/src/action/nav-mobile.js
+++ b/src/action/nav-mobile.js
@@ -72,7 +72,11 @@ class NavAction {
   }
 
   goResetPasswordCurrent() {
-    this._navigate('ResetPasswordCurrent');
+    this._navigate('SetPassword');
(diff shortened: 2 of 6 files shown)


src/action/auth-mobile.js
```diff
@@ -94,7 +96,11 @@ class AuthAction {
       return;
     }
     await this._setToKeyStore(PIN, newPin);
-    await this._generateWalletPassword();
+    if (this._store.walletUnlocked) {
+      this._nav.goResetPasswordSaved();
+    } else {
+      await this._generateWalletPassword();
+    }
```
tanx: Mmmmh. Seems like reusing the `SetPassword` (SetPinView) for mobile is creating a bit of added complexity here. It might be more maintainable to create custom views for mobile and use the same routes as on the desktop (`ResetPasswordCurrent`, `ResetPasswordNew`, `ResetPasswordConfirm`, `ResetPasswordSaved`). I know this would have the downside of creating a bunch of redundancy in view code, but would keep action logic simpler and consistent across platforms. I feel like maintaining view code hasn't been the main source of bugs but rather avoiding weird states in the JS logic. What do you think?

src/action/auth-mobile.js
```diff
@@ -75,6 +75,8 @@ class AuthAction {
     const { auth } = this._store;
     if (auth[param]) {
       auth[param] = auth[param].slice(0, -1);
+    } else if (param === 'newPin' && this._store.walletUnlocked) {
+      this._nav.goSettings();
```
tanx: Same comment here. If we had custom mobile views, we could have a back button in the header to cancel and go back to setting (like we do on desktop).

src/action/auth-mobile.js
valentinewallace: Does this mean the logic of `pushPinDigit`, etc. would be modified to be more like the previous version of this PR: https://github.com/lightninglabs/lightning-app/pull/939/files#diff-783b94fbd9a84833f5931dd9a59e810dR78?

tanx: @valentinewallace sorry I didn't see your comment until now. #939 does look more maintainable in the sense that we create custom views for pin reset. So I think we should probably go ahead and review that one instead. Would you mind rebasing on master and opening a new PR? thanks

valentinewallace: @tanx Sounds good, will do!!

PR #1136, opened 2019-06-04
Update lnd version for safe shutdown fix.

valentinewallace:
We need this update for the safe shutdown fix (#1134) to work.


tanx:
Seems like this is outdated

PR #1148, opened 2019-06-10
Dev/fix slow qr scanner

### LINKED ISSUE #1147: Scanning QR codes is really slow on mobile

tanx:
Closes #1147

### DIFF
--- a/src/component/qrcode-scanner.js
+++ b/src/component/qrcode-scanner.js
@@ -67,7 +67,7 @@ export default class QRCodeScanner extends React.Component {
               scanned ? undefined : (...args) => this.handleCodeScanned(...args)
             }
             barCodeTypes={[BarCodeScanner.Constants.BarCodeType.qr]}
-            style={styles.wrapper}
+            style={StyleSheet.absoluteFill}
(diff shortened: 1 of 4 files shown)


src/component/qrcode-scanner.js
```diff
@@ -41,8 +42,13 @@ export default class QRCodeScanner extends React.Component {
     this.setState({ hasCameraPermission: status === 'granted' });
   }
 
+  handleCodeScanned({ type, data }) {
+    this.setState({ scanned: true });
+    this.props.onQRCodeScanned({ type, data });
```
valentinewallace: Seems you could just pass `data` and not `type`?

tanx: Yeah we only use `data`. But no need to remove type from the QR component. Maybe the component will be used somewhere we need type later.

tanx:
> Fwiw I'm unable to repro this issue 🤔 is it Android specific?

Did you test on mainnet? It wasn't an issue on testnet since querying the route was faster. Likely due to a much smaller network.

PR #1152, opened 2019-06-11
Add parameter to customize Wait screen copy.

### LINKED ISSUE #1038: progress page when sending a payment shows "loading network..."
When a user sends a payment, and it takes a few seconds, then they're presented with the current network loading screen. The text states: "loading network...", which is out of context as the network has already been "loaded", and we're now routing a payment through the network. So it's a matter of using distinct copy when we transition to the loading page due to sending a a payment. Some suggestions for copy:
  * "sending payment..."
  * "routing payment..."

valentinewallace:
Closes #1038.

### DIFF
--- a/src/action/wallet.js
+++ b/src/action/wallet.js
@@ -14,12 +14,13 @@ import { when } from 'mobx';
 import * as log from './log';
 
 class WalletAction {
-  constructor(store, grpc, db, nav, notification) {
+  constructor(store, grpc, db, nav, notification, payment) {
     this._store = store;
     this._grpc = grpc;
     this._db = db;
     this._nav = nav;
     this._notification = notification;
+    this._payment = payment;
   }
 
   //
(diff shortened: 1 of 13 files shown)


src/action/wallet.js
```diff
@@ -245,7 +246,7 @@ class WalletAction {
       this.initResetPassword();
       return this._notification.display({ msg: errorMsg });
     }
-    this._nav.goWait();
+    this._payment.initWaitScreen({ copy: 'Updating password...' });
```
tanx: Why does the wallet need access to the payment action for this? Would probably be cleaner to either:

a. create a wait action
b. let actions call `nav.goWait({ copy: 'Updating password...' }) and set `store.waitScreenCopy` in the nav action

Considering it's only a one liner, I'd lean towards `b.`

valentinewallace: Just remembered this PR exists lol. Updating!

valentinewallace: Updated!

PR #1158, opened 2019-06-11
Fix unified balance

### LINKED ISSUE #1016: Pending-open balance is being added to unifued balance twice

tanx:
Closes #1016

### DIFF
--- a/src/computed/wallet.js
+++ b/src/computed/wallet.js
@@ -17,22 +17,22 @@ const ComputedWallet = store => {
       return store.walletAddress ? `bitcoin:${store.walletAddress}` : '';
     },
     get balanceLabel() {
-      return toAmountLabel(store.balanceSatoshis, store.settings);
+      return toAmountLabel(store.confirmedBalanceSatoshis, store.settings);
     },
     get totalBalanceSatoshis() {
       const {
-        balanceSatoshis: onChainBalanceSatoshis,
+        confirmedBalanceSatoshis: onChainBalanceSatoshis,
         channelBalancePendingSatoshis,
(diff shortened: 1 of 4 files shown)


tanx:
@valentinewallace this on relies on #1156

src/computed/wallet.js
```diff
@@ -17,22 +17,22 @@ const ComputedWallet = store => {
       return store.walletAddress ? `bitcoin:${store.walletAddress}` : '';
     },
     get balanceLabel() {
-      return toAmountLabel(store.balanceSatoshis, store.settings);
+      return toAmountLabel(store.confirmedBalanceSatoshis, store.settings);
     },
     get totalBalanceSatoshis() {
       const {
-        balanceSatoshis: onChainBalanceSatoshis,
+        confirmedBalanceSatoshis: onChainBalanceSatoshis,
```
valentinewallace: Clarifying variable rename :+1:

tanx:
> I think it looks good. So, we were adding the pending force close balance but we should've been adding the total pending close balance?

The problem was that `waiting close` balance was being added in addition to `pending open` balances from autopilot. Since we set autopilot to wait until 1 conf to open new channels this is no longer an issue. This change makes things a lot more resilient in general. This is what @Roasbeef was also mentioning.

PR #1165, opened 2019-06-17
Display payment preimage in transaction details

tanx:


src/view/transaction-detail.js
```diff
@@ -46,6 +46,11 @@ const TransactionDetailView = ({ store, nav }) => (
         <DetailField name="Status">
           {store.selectedTransaction.statusLabel}
         </DetailField>
+        {store.selectedTransaction.preimage ? (
```
valentinewallace: Should we add this to the mobile transaction view as well?

tanx: @valentinewallace good catch! I added it in 6bd1900

PR #1166, opened 2019-06-17
Hide block height in notification bar while syncing

tanx:


valentinewallace:
Ahh, any reason for this? I really like this feature so I know if my wallet is stuck on the last block or still actually syncing.

tanx:
> Ahh, any reason for this? I really like this feature so I know if my wallet is stuck on the last block or still actually syncing.

Dunno. It wasn't really in Lee's original design. It was helpful for testing and debuggin, but not sure we want it in a production app.

PR #1168, opened 2019-06-19
Dev/secure random bytes

tanx:
This is once again integrated in `v0.1.1`. Let's see if we run into issues again on the Pixel 3. It tested without issues on Nexus 5X and Moto G7. I definitely feel better shipping with this and potentially crashing on the Pixel 3 than shipping insecure wallet encryption on all devices.

### DIFF
--- a/mobile/android/app/build.gradle
+++ b/mobile/android/app/build.gradle
@@ -104,6 +104,7 @@ apply from: 'expo.gradle'
 // WHEN_PREPARING_SHELL_REMOVE_TO_HERE
 
 dependencies {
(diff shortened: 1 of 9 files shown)


halseth:
How is it different this time?

valentinewallace:
Breaks travis?

tanx:
> Breaks travis?

@valentinewallace fixed this in the current mobile release branches.

> How is it different this time?

@halseth nothing. I think something else caused the Android app to crash. Not this patch. The JS warning comes up every time. even on non-crashes.

PR #1190, opened 2019-06-20
MXN amounts; depends on MXN ticker to be included on API response

mautematico:
It depends on MXN rates being included on API response; may be we could use Bitso's:
https://api.bitso.com/v3/ticker/?book=btc_mxn

I'll be glad to send a PR to API repo if you point me to where is it.


alexbosworth:
MXN is included in API response now

PR #1207, opened 2019-06-21
Dev/blank seed

### LINKED ISSUE #1198: Seed is empty for some users (can be solved by restarting)

valentinewallace:
Closes #1198.


src/action/wallet.js
```diff
@@ -320,7 +322,7 @@ class WalletAction {
    * @return {undefined}
    */
   initRestoreWallet() {
-    this._store.seedMnemonic = Array(24).fill('');
+    this._store.restoreSeedMnemonic = Array(24).fill('');
```
tanx: So the seed words were empty because the navigated to `RestoreWallet` after the seed had been set by `wallet.generateSeed`? Couldn't we just call `wallet.generateSeed()` in `wallet.initSeed()`?

valentinewallace: That works, fixed!

Note that we still need [this](https://github.com/lightninglabs/lightning-app/blob/master/src/action/wallet.js#L140) call to `generateSeed()` so the `generateSeed()` call in `initSeed()` will be redundant most of the time.

tanx: Thanks for looking into this variant. I just noticed that we use `wallet.generateSeed` to check if a wallet already exists. Calling it twice without handling the error would create another potential error case. I think you made the right call in the first version of your PR to add `store.restoreSeedMnemonic`. I reopened that variant in https://github.com/lightninglabs/lightning-app/pull/1247 and am closing this one.

Issue #1180, opened 2019-06-19
Set non-standard port for lnd on Android to prevent app from crash in case another wallet is open


tanx:
Set `--nolisten`

Roasbeef:
An alternative solution proposed in: #1184 (basically just disable listening all together). For the testnet vs mainnet app, we can use different RPC ports via these two options: 
```
      --rpclisten=                                            Add an interface/port/socket to listen for RPC connections
      --restlisten=                                           Add an interface/port/socket to listen for REST connections
```

mautematico:
For instance, ligtning-app and @ACINQ/eclair-mobile are unable to run at the same time.

tanx:
thanks @Roasbeef !

PR #1209, opened 2019-06-22
mobile: disable listening on Android+iOS

### LINKED ISSUE #1184: [alpha]mobile: app should disable listening
Reported by many users: if they have another LN app installed on their phone, then ours will crash once it finishes syncing. This is due to the fact that the other applications are tyring to listen on the default p2p port (9735). As this is a mobile app, we don't really need to be listening at all. As a result, we should start `lnd` with the following flag: 
```
      --nolisten                                              Disable listening for incoming peer connections
```

Roasbeef:
In this commit, we disable listening on Android and iOS in order to
ensure that we don't conflict with any other LN apps the user may have
that are also attempting to listen on port 9735.

Fixes #1184.

### DIFF
--- a/mobile/android/app/src/main/assets/lnd.conf
+++ b/mobile/android/app/src/main/assets/lnd.conf
@@ -2,6 +2,7 @@
 debuglevel=info
 no-macaroons=1
 maxbackoff=2s
+nolisten=1
 
 [Routing]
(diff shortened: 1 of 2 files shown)


tanx:
utACK

tanx:
Also closes #1180

Issue #1174, opened 2019-06-19
iOS: Creating a new wallet results in a screen with blank seed words


alexbosworth:
### Description

When I try to make a wallet on iOS it results in blank seed words.

### Steps to reproduce the behavior

1. Install Lightning version 0.1.1 (6) on testflight
2. Go to 'generate new seed'
3. See empty seed words where the seed words should be

### Expected behavior

Seed words should appear

### Actual behavior

Seed words don't appear on any of the seed word screens

### Workaround

Killing and restarting the App makes the seed words come back

alepop:
I got the same issue. Step to reproduce:
- Open app
- Click "restore from seed"
- Go back
- Click "generate new seed"

valentinewallace:
I think this issue was resolved by the merge of https://github.com/lightninglabs/lightning-app/pull/1247

tanx:
> I think this issue was resolved by the merge of #1247

Yep

PR #1212, opened 2019-06-24
fix empty seedMnemonic in initSeed

### LINKED ISSUE #1206: [Android] Empty seed words when creating a wallet
### Description
Android shows empty seed words reproducibly, like #1174 

Choose create a new wallet on App startup. Go back. Choose recover an existing wallet. Go back. Choose create a new wallet. Seed words will be empty.

### Expected behavior
Create a new random seed whenever Create a new wallet is pressed on startscreen.

### Actual behavior
See above.

### Screenshot and/or logs if applicable
Screenshot_20190620-140126_Lightning

MichelKansou:
Origin of the bug #1206 & #1174 the seed is cleared if the user select recover an existing wallet then navigate back to home/Main screen and select generate a new wallet, so I suggest to check in `initSeed()` if the seedMnemonic is empty before navigating to generate a new wallet screen!

### DIFF
--- a/src/action/wallet.js
+++ b/src/action/wallet.js
@@ -279,8 +279,8 @@ class WalletAction {
   }
 
   /**
-   * Initialize the seed flow by navigating to the proper next view and
-   * resetting the current seed index.
+   * Check the existing mnemonic seed and Initialize the seed flow by navigating to
+   * the proper next view and resetting the current seed index.
    * @return {undefined}
    */
   async initSeed() {
--- a/src/action/wallet.js
+++ b/src/action/wallet.js
@@ -283,7 +283,10 @@ class WalletAction {
    * resetting the current seed index.
    * @return {undefined}
    */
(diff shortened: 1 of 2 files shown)


tanx:
We already `await` in `wallet.init()`. This just calls `generateSeed()` again. I'm assuming just decreases the likelyhood of the underlaying race condition.

tanx:
Gonna close this until we figure out what's causing the race condition.

MichelKansou:
@tanx sorry but I don't see when the `wallet.init()` is called anywhere after `wallet.initPrevRestorePage()` the seed is cleared because when user navigate to `RestoreSeedView` after calling `initRestoreWallet`
```
 initRestoreWallet() {
    this._store.seedMnemonic = Array(24).fill('');
    this._store.wallet.restoreIndex = 0;
    this._nav.goRestoreSeed();
  }
```

PR #1239, opened 2019-07-02
Initial commit of bare project

tanx:
@valentinewallace this PR is just the generated react-native project for the expo bare build. My suggested way to integrate these changes would be to:

1. Review (not yet merge) this PR which adds the template under `/mobile-bare`
2. Review the changes to the template in a separate PR to this branch
3. `git rm mobile` and `git mv mobile-bare mobile` to replace `/mobile`
4. Merge the final PR which completely replaces the old expo build in `/mobile`


mobile-bare/.buckconfig
```diff
@@ -0,0 +1,6 @@
+
```
valentinewallace: Pains me to add unused config files for expokit's dependencies, are they necessary for the project to run/compile? :stuck_out_tongue: (noticed this one + the watchman config are empty, plus some other files that are blank except for comments)

tanx: > Pains me to add unused config files for expokit's dependencies, are they necessary for the project to run/compile?

I think this file is generated even in a regular react native project. Why do you think it's unused?

PR #1253, opened 2019-07-10
Mobile/bare react native forgot pin

### LINKED ISSUE #1227: Add "forgot password" button on PIN screen in case user is locked out
This will init wallet restore from seed and allow users to restore their wallet and channels from the SCB on disk.

so hot swap flow:
  * shut down lnd (stop rpc)
  * delete all but `channel.db` (using https://github.com/itinance/react-native-fs)
  * restart lnd 
  * do regular `InitWallet` as normal

Pin - Mobile

Pin - Mobile - Recovery

tanx:
Closes #1227 

Don't merge


src/action/wallet.js
```diff
@@ -265,6 +266,7 @@ class WalletAction {
    */
   async initWallet({ walletPassword, seedMnemonic, recoveryWindow = 0 }) {
     try {
+      await Promise.all([this.deleteDB('testnet'), this.deleteDB('mainnet')]);
```
valentinewallace: Just a note: if you're going through this PR's reset flow, but you use the wrong seed, wallet initialization will fail every time. `Wallet initialization failed: undefined`

src/action/auth-mobile.js
```diff
@@ -258,6 +270,41 @@ class AuthAction {
...
+          text: action,
+          onPress: () => this._initRestoreWallet(),
+        },
+        { text: cancel, style: 'cancel' },
+      ]);
+    }
+  }
+
+  _initRestoreWallet() {
+    this._store.settings.restoring = true;
```
valentinewallace: minor: could move this line 305 into `wallet.initRestoreWallet`, then we can remove this helper

test/unit/action/wallet.spec.js
```diff
@@ -303,7 +308,31 @@ describe('Action Wallet Unit Tests', () => {
...
+      );
+      expect(store.walletUnlocked, 'to be', true);
+      expect(grpc.sendUnlockerCommand, 'was called with', 'InitWallet', {
+        walletPassword: Buffer.from('baz', 'utf8'),
+        cipherSeedMnemonic: ['foo'],
+      });
+      expect(nav.goSeedSuccess, 'was called once');
+    });
+
+    it('should not delete wallet if RNFS not supported', async () => {
```
valentinewallace: Curious in what case `RNFS` wouldn't be supported? Does this mean if `wallet.db` doesn't exist?

src/action/auth-mobile.js
```diff
@@ -258,6 +270,41 @@ class AuthAction {
...
+  // Help / Restore actions
+  //
+
+  askForHelp() {
+    const message =
+      "If you have forgotten your PIN, or you're locked out of your wallet, you can reset your PIN with your Recovery Phrase.";
+    const action = 'Recover My Wallet';
+    const cancel = 'Cancel';
+    if (this._Platform.OS === 'ios') {
+      this._ActionSheetIOS.showActionSheetWithOptions(
```
valentinewallace: Looks slick!

src/action/wallet.js
tanx: @Roasbeef any ideas?

tanx:
> We should figure out why travis is unhappy though 🤔

Looks like travis just got caught up for some reason. It works now.

src/action/auth-mobile.js
tanx: Thought about that too. But seems a bit unclean since we set that before `wallet.initRestoreWallet()` in the select-seed screen.

test/unit/action/wallet.spec.js
tanx: That's handled in https://github.com/lightninglabs/lightning-app/pull/1253/files#diff-fbb9a2d7e2e2b04543043634746adfe1R292

src/action/wallet.js
tanx: > Just a note: if you're going through this PR's reset flow, but you use the wrong seed, wallet initialization will fail every time. `Wallet initialization failed: undefined`

@valentinewallace I added a try/catch case and specific error message for this case. It's in the current testnet build.

PR #1258, opened 2019-07-11
Select autopilot during wallet setup.

### LINKED ISSUE #1167: Let users choose if autopilot is enabled during setup
paste

valentinewallace:
Closes #1167. Tested in an iOS build. I didn't add the feature to desktop.

### DIFF
--- a/src/component/setting.js
+++ b/src/component/setting.js
@@ -85,10 +85,19 @@ const iStyles = StyleSheet.create({
   item: {
     height: 60,
   },
-  name: {
+  left: {
     flex: 1,
+  },
+  name: {
     color: color.grey,
     fontSize: font.sizeSub,
+  },
+  copy: {
+    width: '80%',
+    fontSize: font.sizeS,
+    lineHeight: font.lineHeightS,
+    color: color.greyListLabel,
+    opacity: 0.74,
   },
   lbl: {
     fontSize: font.sizeS,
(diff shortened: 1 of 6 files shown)


src/view/select-atpl-mobile.js
```diff
@@ -0,0 +1,91 @@
...
+        Automatically set up channels?
+      </CopyOnboardText>
+      <Text style={styles.copyTxt}>
+        {
+          "If you want to avoid manual channel creation, the app can allocate funds to channels for you. It's called autopilot."
+        }
+      </Text>
+      <View style={styles.list}>
+        <SettingItem
+          name="Use autopilot (recommended)"
```
valentinewallace: I added the word "recommended"... Feel free to remove if you prefer.

src/component/setting.js
```diff
@@ -100,9 +109,22 @@ const iStyles = StyleSheet.create({
...
+  children,
+}) => (
+  <ListItem
+    style={[iStyles.item, copy ? { height: 80 } : null]}
+    onSelect={onSelect}
+  >
+    <View style={iStyles.left}>
+      <Text style={iStyles.name}>{name}</Text>
+      {copy ? <Text style={iStyles.copy}>{copy}</Text> : null}
+    </View>
```
tanx: Would probably be better to create a separate `SettingCopyItem` component than trying to squeeze both use cases into one component. That would also allow us to set the margins/padding according to the design, as they look pretty different right now.

Design:


Implementation:

src/view/select-atpl-mobile.js
tanx: makes sense 👍

valentinewallace:
Updated @tanx :)

src/component/setting.js
valentinewallace: Thanks, must've been in a hurry hah

PR #1267, opened 2019-07-18
Export logs

### LINKED ISSUE #1109: Allow download of complete log file in mobile app
Add a share button on the upper right corner of the CLI/LOGs view header using the react-native `share` api and the react-native-fs `readFile` api.

See: https://facebook.github.io/react-native/docs/share#share

See: https://github.com/itinance/react-native-fs#readfilefilepath-string-encoding-string-promisestring

The log file can be found under: https://github.com/lightninglabs/lightning-app/blob/master/src/action/wallet.js#L304

Button should appear in the upper right of the header:

image

wbobeirne:
Closes #1109.

### What This Does

Adds a share button to the top right of the logs on mobile that loads the log full log file and pops up the device's native share dialog to be able to send it somewhere else (Most likely to save it as a file or send it in an email.)

For now, the button just doesn't show on the desktop electron app, but maybe a good follow up would be to have this button open the folder the log file is in so that they could easily upload it.

### Steps to Test

1. Open the app on mobile and go to your logs, confirm you see a share button in the top right
2. Press the share button, confirm you are able to export the logs to your preferred app
3. Open the app on desktop and go to your logs, confirm you do not see a share button in the top right

## Screenshots

log-export


src/action/log.js
```diff
@@ -49,6 +50,29 @@ export function error(...args) {
...
+ */
+export function getLogs() {
+  if (!_FS) {
+    return Promise.reject(
+      'Cannot get logs when no filesystem is available in action/log.js'
+    );
+  }
+  return _FS.readFile(getLogPath(_store.network), 'utf8');
+}
+
```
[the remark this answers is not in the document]
wbobeirne: Added some tests with and without the required constructor args. Let me know if the way I stubbed RNFS and Share was what you normally do.

src/view/cli.js
```diff
@@ -19,16 +20,37 @@ const styles = StyleSheet.create({
...
+    try {
+      const logs = await getLogs();
+      await Share.share({
+        title: 'Lightning App logs',
+        message: logs,
+      });
+    } catch (err) {
+      error(err);
+    }
+  }
```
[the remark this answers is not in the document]
wbobeirne: Good call, moved it out to a `shareLogs` function exported from `actions/log.js`

src/action/log.js
```diff
@@ -49,6 +51,49 @@ export function error(...args) {
...
+/**
+ * Retrieves the entire LND log file as a string.
+ * @return {Promise}
+ */
+export function getLogs() {
+  if (!_FS) {
+    return Promise.reject(
+      new Error('Cannot get logs with no FS in action/log.js')
+    );
```
tanx: If we declare `async function getLogs` we be can throw an Error here directly. No need to wrap in `Promise.reject()` in that case.

src/action/log.js
```diff
@@ -49,6 +51,49 @@ export function error(...args) {
...
+  if (!_FS) {
+    throw new Error('Cannot get log path with no FS in action/log.js');
+  }
+  const lndDir = _FS.DocumentDirectoryPath;
+  return `${lndDir}/logs/bitcoin/${network}/lnd.log`;
+}
+
+/**
+ * Retrieves the entire LND log file as a string.
+ * @return {Promise}
```
tanx: `@return {Promise<string>}`

src/action/log.js
```diff
@@ -49,6 +51,49 @@ export function error(...args) {
...
+
+/**
+ * Shares the log file using whatever native share function we have.
+ * @return {Promise}
+ */
+export async function shareLogs() {
+  if (!_Share) {
+    return Promise.reject(
+      new Error('Cannot share logs with no Share in action/log.js')
+    );
```
tanx: Same here ... no need to wrap in `Promise.reject` due to `async function shareLogs`. Just do a `throw new Error(...)`.

src/action/log.js
```diff
@@ -49,6 +51,49 @@ export function error(...args) {
...
+  }
+  const logs = await getLogs();
+  return _Share.share({
+    title: 'Lightning App logs',
+    message: logs,
+  });
```
tanx: We should wrap the complete contents of `shareLogs` in a try/catch and log the error here with `this.error(err)`. Then we don't need to wrap the function call in a try/catch in the view. That way we could keep the view react component purely functional.

src/action/log.js
```diff
@@ -59,10 +104,12 @@ function pushLogs(message) {
 }
 
 class LogAction {
-  constructor(store, ipc, printErrObj = true) {
+  constructor(store, ipc, printErrObj = true, FS, Share) {
```
tanx: 👍

src/view/cli.js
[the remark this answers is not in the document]
tanx: As mentioned above, we should move this try/catch block into `shareLogs` itself and call `onPress={() => log.shareLogs()}`. Then we can keep the react component purely functional as before and prevent breakout into the `class` syntax.

src/view/cli.js
```diff
@@ -19,16 +20,33 @@ const styles = StyleSheet.create({
   },
 });
 
-const CLIView = ({ store, nav }) => (
```
tanx: We should inject the `log` instance of the LogActions as a prop into this view via `({ store, nav, log })`. And inject `log` in https://github.com/lightninglabs/lightning-app/blob/1ee33ed1651bd84a87174a5ebc3c16b27c0e030f/src/view/main-mobile.js#L140

test/unit/action/log.spec.js
```diff
@@ -10,6 +10,14 @@ describe('Action Logs Unit Tests', () => {
...
   let ipcRenderer;
 
+  const RNFS = {
+    DocumentDirectoryPath: '/foo/bar',
+    readFile: () => Promise.resolve('logs'),
+  };
+  const Share = {
+    share: () => Promise.resolve(true),
+  };
+
```
tanx: Stubs should be declared via `let` here and initialized using `sinon.stub().resolves(true)`  in the `beforeEach` block below. That way all test case can work with a new stub instance and check how often it was called.

src/action/log.js
wbobeirne: Are there any instances in which we may want to have custom logic when it throws? (E.g. popping up an error message to the user?) I typically avoid eating errors in case it needs to be handled.

src/view/cli.js
wbobeirne: Right now all of the log action's functions are standalone and individually exported & imported. Should they all be moved into the Log class and called like that?

wbobeirne:
* Removed unnecessary `Promise.reject`'s
* Made `shareLogs` a self-handling component so that it doesn't require catching errors
* Added sinon stubs to tests and check that the relevant functions get called

The only outstanding issue is the passing of `log` as a prop, let me know how you feel about that one and if it needs changes.

tanx:
> * Removed unnecessary `Promise.reject`'s
> * Made `shareLogs` a self-handling component so that it doesn't require catching errors
> * Added sinon stubs to tests and check that the relevant functions get called

@wbobeirne awesome! Thanks for the updates.

> The only outstanding issue is the passing of `log` as a prop, let me know how you feel about that one and if it needs changes.

Indeed. It doesn't seem to be so simple since we don't actually use this action that way for the other functions. I can look into this. No need to dig further into it on your end.

Thanks so much for the PR 👍

tanx:
@wbobeirne I went ahead and moved the file I/O operations into its own action module. That allowed us to inject `file` via a prop into the cli view like the other actions. Other than that great work!

Closing this PR in favor of https://github.com/lightninglabs/lightning-app/pull/1270

PR #1284, opened 2019-08-05
password validation on the first step

### LINKED ISSUE #1279: Invalid password should not be able to move to next step
### Description

When creating a wallet, on the "Set a Password" step, you can click "Next" to go to the next step ("Confirm Password") without the password being acceptable.

1. Create a new wallet.
2. On the "Set a Password" step, click "Next" without typing any password or typing an invalid password (like "test").

### Expected behavior

It goes to the "Confirm Password" step, without any error message.

### Actual behavior

It should remain on the same step and give a warning to the user.

skv-headless:
closes #1279


src/action/wallet.js
```diff
@@ -201,13 +201,23 @@ class WalletAction {
     this.initSetPassword();
   }
 
+  checkNewPassword() {
```
skv-headless: I can add method documentation if PR in general is good

PR #1289, opened 2019-08-07
ios native: pass localizedDescription as error message

halseth:
It will also be part of the passed NSError, but this makes the error
message explicit.

### DIFF
--- a/src/action/payment.js
+++ b/src/action/payment.js
@@ -247,7 +247,7 @@ class PaymentAction {
       this._nav.goPayBitcoinConfirm();
     } catch (err) {
       this._notification.display({
-        msg: `Fee estimation failed: ${err.details}`,
+        msg: `Fee estimation failed: ${err}`,
         err,
       });
     }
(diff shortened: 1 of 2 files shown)


src/action/payment.js
```diff
@@ -247,7 +247,7 @@ class PaymentAction {
       this._nav.goPayBitcoinConfirm();
     } catch (err) {
       this._notification.display({
-        msg: `Fee estimation failed: ${err.details}`,
+        msg: `Fee estimation failed: ${err}`,
```
tanx: @halseth can we change it so that `err` is an object that has the string on a `message` attribute so that we can access it like a normal JS Error like so: `err.message`?

halseth: I can't really set anything, it is react-native that creates this error object.

tanx: To clarify, we're not getting an error object here. `err` is a string.

halseth: Ok, I guess that's what React Native throws then :)

tanx:
Closing in favor of #1292

PR #1295, opened 2019-08-20
Dev/fix home balance font size

### LINKED ISSUE #879: Resize balance label to fit screen width on mobile (home screen)

tanx:
Closes #879 

Used Lee's updated design to allow `Text.adjustsfontsizetofit`:

image

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -94,7 +94,7 @@ export class Store {
       settings: {
         unit: DEFAULT_UNIT,
         fiat: DEFAULT_FIAT,
-        displayFiat: true,
+        displayFiat: false,
         exchangeRate: {},
         restoring: false,
         autopilot: true,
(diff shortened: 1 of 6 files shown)


src/store.js
```diff
@@ -94,7 +94,7 @@ export class Store {
       settings: {
         unit: DEFAULT_UNIT,
         fiat: DEFAULT_FIAT,
-        displayFiat: true,
+        displayFiat: false,
```
valentinewallace: So bold

tanx:
> tACK, works great!
> 
> I'm getting this warning, but I can't figure out why because it seems like the `prop` is a string:

@valentinewallace good catch. Fixed this in https://github.com/lightninglabs/lightning-app/pull/1295/commits/8305fffc380bf4de5585baab296009c419a23ea0

PR #1308, opened 2019-09-21
Remove 'gomobile init' from README

joelklabo:
`gomobile init` is no longer necessary to complete setup. Shows an error.


tanx:
Gomobile docs still say it's required: https://godoc.org/golang.org/x/mobile/cmd/gomobile

You need to upgrade to go `1.13` for the current gomobile.

Issue #1269, opened 2019-07-22
pressing back in the pin entry screen leaves the user stuck in limbo


djpnewton:
if you press back in the pin entry screen you are stuck in a splash screen limbo

PR #1321, opened 2019-11-07
Fix splash screen hang when press back

BSathvik:
- Fix issue #1269


BSathvik:
Is this good to go @Roasbeef ?

PR #1325, opened 2020-01-09
Update mobile/README.md on where to get falafel.

emiljoha:
github.com/halseth/falafel is as of this writing on version 0.5 while the lightninglabs repo is on 0.6 and the build requires version 0.6. Looks like halseth is doing the commits in the lightninglabs repo as well so not really sure what is going on.


emiljoha:
Turns out only lnd release v0.8.2 is supported at the moment as the latest lnd release has a breaking api change. lnd v0.8.2 has falafel v0.5 as dependency which is what github.com/halseth/falafel is at. So all is well, kind of.

PR #1327, opened 2020-01-12
Redirect navigation flow for 0 value invocies. Fixes #991

### LINKED ISSUE #991: app doesn't support sending to 0 value invoices
Seems people are into sending tips over LN these days. They usually do so by giving out a "zero valued" input which looks like this when decoded: 
```
top@kek9000:~/gocode/src/github.com/lightningnetwork/lnd# lncli -n testnet  decodepayreq lntb1pwxmh7lpp5mnsmlk7rhtj6fuj362ff5xwtxjssczrt84q04pucjnj407046uqsdqqcqzysm0l4rl86nuv94ahea3rv4fqsqymqm4xmaq5xthe6nnnnvcc227srmufarxjw7euhtj8lftqtr8pk5xaga50aheey8yjzdlyc2tv4pkcp77yrhl
{
    "destination": "0270685ca81a8e4d4d01beec5781f4cc924684072ae52c507f8ebe9daf0caaab7b",
    "payment_hash": "dce1bfdbc3bae5a4f251d2929a19cb34a10c086b3d40fa879894e557f9f5d701",
    "num_satoshis": "0",
    "timestamp": "1550704607",
    "expiry": "3600",
    "description": "",
    "description_hash": "",
    "fallback_addr": "",
    "cltv_expiry": "144",
    "route_hints": [
    ]
}
```

So one the decoder side, we can detect this at it has no true value: 
```
    "num_satoshis": "0",
```

When we try to pay an invoice like this into the UI atm we see something like this 
screen shot 2019-02-20 at 3 21 13 pm

When the user clicks to pay, then it gets rejected as `lnd` detects there's no value attached to the send request. 

If we wan to support this use case, we'll need to:
  1. Detect this special type of invoice while we're validating the user data
  2. Modify the currently static "send payment" UI fragment to be conditionally dynamic if there's a zero valued invoice. In this case, the user would get some sort of text saying they should specify the amount since it's a tip/donation. 
  3. Take that value and specify it in the `SendRequest` proto _along side_ the invoice. Here's an example of how we do it in lnd/Go: https://github.com/lightningnetwork/lnd/blob/master/cmd/lncli/commands.go#L2255. Notice that we set _both_ the invoice and the amount. If the amount is zero, then lnd will just use the invoice. But if it's set, then lnd knows what do to with it all.

emiljoha:
- [x] Functionality in desktop. (Tested on Linux) 
- [x] Functionality in mobile (Tested on android)
- [x] Fix: Background on new view does not properly match views before and after. 
- [x] Make sure text content and placement is consitent with rest of the app.
- [x] Fix: Input field on Desktop becomes 0500 when trying to write 500.
- [x] Error handling. Make sure all error scenarios are considered and handled properly. (Note: My experience in UI coding is limited but I am currently convinced this new feature will handle errors gracefully)

more-or-less-polished


emiljoha:
Basic idea of the UX.
zero-amount-invoice-ux

emiljoha:
Done

[x] Functionality in desktop. (Tested on Linux) 

[x] Functionality in mobile (Tested on android)

Needs to be fixed: 

[ ] Background on new view does not properly matchview before and after. 

[ ] Error handling. Make sure all error scenarios are considered and handled properly. 

[ ] (Optional) Receiving zero value invoices give notice of receiving 0 sats. Optional as I am hard pressed to come up with a common use-case for sending zero value invoices on mobile .

emiljoha:
any-amount-invoice-payment

src/action/payment.js
```diff
@@ -165,14 +165,35 @@ class PaymentAction {
       return this._notification.display({ msg: 'Enter an invoice or address' });
     }
     if (await this.decodeInvoice({ invoice: this._store.payment.address })) {
-      this._nav.goPayLightningConfirm();
+      if (
+        this._store.payment.amount === '0' ||
+        this._store.payment.amount === 0
```
bolatovumar: is there a need for checking for both `0` and `'0'`? I feel like it should always just be a value of one type. If that's not the case then it should be cast to one type first before checking.

jsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
```
bolatovumar: I'm not sure this file really needs to be in this PR. It's useful but has nothing to do with this particular feature. I think this should be added in a follow-up PR.

src/action/payment.js
```diff
@@ -199,6 +220,27 @@ class PaymentAction {
...
   }
 
+  /**
+   * Estimate the lightning transaction fee using the queryRoutes grpc api
+   * after which the fee is set in the store.
+   * @param  {number} options.satAmt      The amount to be payed in satoshis
+   * @return {Promise<undefined>}
+   */
+  async reEstimateLightningFee() {
+    try {
```
bolatovumar: I wonder if the error from this try/catch block should be propagated further from this method so that the UI can display some useful info to the use and try to recover rather than just having a log message.

src/action/payment.js
```diff
@@ -321,7 +363,8 @@ class PaymentAction {
   }
 
   /**
-   * Send the amount specified in the invoice as a lightning transaction and
+   * Send the amount specified in the payment.amount. If zero use the amount
```
bolatovumar: Wonky description here: `If zero use the amount specified in the invoice as a lightning transaction and display the wait screen while the payment confirms.` doesn't really make sense.

src/action/payment.js
emiljoha: Not really, I should :tm: be '0' its just me being defensive. Will fix.

jsconfig.json
emiljoha: Not sure why this is there or what it does... Probably got created when I was trying to set up the project/IDE. Will remove from this PR. Not really sure what it does but I can add it in a follow up PR if It would be useful.

src/action/payment.js
emiljoha: Thinking about it, that exception will only be thrown if we do not have a valid lightning invoice in the address field. That is dependent on other functions being called before this one. That kind of subtle and implicit dependencies are dangerous, IMHO. I have tried to make the expectations of the method more clear both in code and documentation. 

PS: I am used to writing functional Typescript code in my day job, so all this state and no typing is making me a bit uncomfortable.

src/action/payment.js
emiljoha: Will try to make it more clear.

src/action/payment.js
emiljoha: Fixed in d1c9dfe8

jsconfig.json
emiljoha: Fixed in 86f0fa3e
Follow up PR: #1331

src/action/payment.js
emiljoha: Fixed in 93227257 and 51413f29

src/action/payment.js
emiljoha: Fixed in 55985e7d

emiljoha:
@bolatovumar Thank you for you comments! Have tried to address the issues.

jsconfig.json
bolatovumar: @emiljoha it's used by VSCode to configure the built-in JS language service. See here for details: https://code.visualstudio.com/docs/languages/jsconfig

emiljoha: Ah! That makes a lot of sense. Thanks!

emiljoha:
@Roasbeef  (Pinging you as you where the originator of the issue and the only one that have merged commits into to this repository for the last couple of months.)

The change is IMHO ready and has been reviewed by @bolatovumar. Please advice if there are any further steps needed to be taken in order to move forward in solving this issue.

bolatovumar:
I have reviewed the PR and it looks good to me. I have not tested it, however.

PR #1331, opened 2020-02-01
Add jsconfig.

emiljoha:


bolatovumar (commented):
Shouldn't `node_modules` folder be added to `exclude` property? See here: https://code.visualstudio.com/docs/languages/jsconfig#_best-practices

emiljoha:
@bolatovumar  Excluding node_modules folder sounds very reasonable and is supported by the vscode docs. Fixed in 4125e9ad

